PR #14, opened 2018-03-07
Tests for OpenVPN monitor.

ababo:
This change fully resolves **BV-92**.

### DIFF
--- a/vpn/mon/monitor.go
+++ b/vpn/mon/monitor.go
@@ -83,6 +83,12 @@ func (m *Monitor) requestClients() error {
 	return nil
 }
 
+// Monitor errors.
+var (
+	ErrComm          = errors.New("openvpn communication error")
+	ErrTooOldOpenVPN = errors.New("too old OpenVPN version")
+)
+
 const (
 	prefixClientListHeader  = "HEADER,CLIENT_LIST,"
 	prefixClientList        = "CLIENT_LIST,"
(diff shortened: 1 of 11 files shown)


vpn/mon/monitor.go
```diff
@@ -83,6 +83,12 @@ func (m *Monitor) requestClients() error {
 	return nil
 }
 
+// Monitor errors.
+var (
+	ErrComm          = errors.New("openvpn communication error")
+	ErrTooOldOpenVPN = errors.New("too old OpenVPN version")
```
furkhat: Do we need to export this names?

vpn/util/misc.go
```diff
@@ -19,8 +20,16 @@ func FindCurrentSession(db *reform.DB, ch string) (string, error) {
...
-// consumed VPN traffic and received payments.
-func HasPositiveBalance(db *reform.DB, ch string) (bool, error) {
-	return true, nil
+// ChannelOpen checks whether a given state channel is open. For VPN service
+// that means it allows to connect and to consume traffic.
+func ChannelOpen(db *reform.DB, ch string) (bool, error) {
+	var open bool
+	if err := db.QueryRow(`
+		SELECT state = $1
+		  FROM channels
```
furkhat: May be we should move this to some other package?

ababo: Let's leave it here, but will keep in mind that this functionality can be reused by non-VPN functionality to be ready to move it.

PR #15, opened 2018-03-07
fix readme.md

drew2a:
@ababo @HaySayCheese  @furkhat I marked blocks that need your attention as @dappctrl_team 
:)


ababo:
Makes sense to add a command to run tests:
```bash
go test $DAPPCTRL/... -config=$DAPPCTRL_DIR/dappctrl-test.config.json
```

PR #17, opened 2018-03-09
payment(balance proof) receiver server

furkhat:
Payment receiver server implemented as a distinct server on a separate `http.ServeMux`. Tests are agains real postgres db using `httptest`.

### DIFF
--- a/dappctrl-test.config.json
+++ b/dappctrl-test.config.json
@@ -2,12 +2,18 @@
     "DB": {
         "Conn": {
             "dbname": "dappctrl",
-            "sslmode": "disable"
+            "sslmode": "disable",
+            "user": "postgres"
         }
     },
 
     "Log": {
         "Level": "info"
+    },
+
+    "PaymentServer": {
+        "Addr": "localhost:9000",
+        "TLS": null
     },
 
     "TestData": {
--- a/data/schema.go
+++ b/data/schema.go
@@ -72,7 +72,7 @@ type Channel struct {
 	Agent            string `reform:"agent"`
 	Client           string `reform:"client"`
 	Offering         string `reform:"offering"`
-	Block            uint   `block`
+	Block            uint   `reform:"block"`
 	State            string `reform:"state"`
 	TotalDeposit     string `reform:"total_deposit"`
 	ClosedDeposit    string `reform:"closed_deposit"`
--- a/data/test.go
+++ b/data/test.go
@@ -3,9 +3,14 @@
 package data
 
 import (
+	"crypto/ecdsa"
+	cryptorand "crypto/rand"
 	"log"
+	"math/rand"
 
-	"gopkg.in/reform.v1"
+	"github.com/ethereum/go-ethereum/common/number"
+	"github.com/ethereum/go-ethereum/crypto"
+	reform "gopkg.in/reform.v1"
 
 	"github.com/privatix/dappctrl/util"
 )
@@ -25,4 +30,45 @@ func NewTestDB(conf *DBConfig, logger *util.Logger) *reform.DB {
 		log.Fatalf("failed to open db connection: %s\n", err)
 	}
 	return db
+}
+
+// NewTestSubject returns new subject
+func NewTestSubject() *Subject {
+	priv, _ := ecdsa.GenerateKey(crypto.S256(), cryptorand.Reader)
+	b := crypto.FromECDSA(priv)
+	privB64 := FromBytes(b)
+	priv, _ = crypto.ToECDSA(b)
+	pub := FromBytes(
+		crypto.FromECDSAPub(&priv.PublicKey))
+	return &Subject{
+		ID:         util.NewUUID(),
+		PrivateKey: &privB64,
+		PublicKey:  pub,
+	}
+}
+
+// NewTestOffering returns new offering.
+func NewTestOffering(agent *Subject) *Offering {
+	return &Offering{
+		ID:      util.NewUUID(),
+		Agent:   agent.ID,
+		Service: ServiceVPN,
+		Supply:  1,
+	}
+}
+
+// NewTestChannel returns new channel.
+func NewTestChannel(agent, client *Subject, offering *Offering,
+	balance, deposit int64, state string) *Channel {
+	return &Channel{
+		ID:             util.NewUUID(),
+		Agent:          agent.ID,
+		Client:         client.ID,
+		Offering:       offering.ID,
+		Block:          uint(rand.Intn(99999999)),
+		State:          state,
+		TotalDeposit:   FromBytes(number.Big(deposit).Bytes()),
+		ClosedDeposit:  FromBytes(number.Big(0).Bytes()),
+		ReceiptBalance: FromBytes(number.Big(balance).Bytes()),
+	}
 }
--- a/main.go
+++ b/main.go
@@ -2,18 +2,21 @@ package main
 
 import (
 	"flag"
+	"log"
+
 	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/payment"
 	"github.com/privatix/dappctrl/util"
 	vpnmon "github.com/privatix/dappctrl/vpn/mon"
 	vpnsrv "github.com/privatix/dappctrl/vpn/srv"
-	"log"
 )
 
 type config struct {
-	DB         *data.DBConfig
-	Log        *util.LogConfig
-	VPNServer  *vpnsrv.Config
-	VPNMonitor *vpnmon.Config
+	DB            *data.DBConfig
+	Log           *util.LogConfig
+	PaymentServer *payment.Config
+	VPNServer     *vpnsrv.Config
+	VPNMonitor    *vpnmon.Config
 }
 
 func newConfig() *config {
@@ -60,5 +63,7 @@ func main() {
 			mon.MonitorTraffic())
 	}()
 
-	<-make(chan bool)
+	pmt := payment.NewServer(conf.PaymentServer, logger, db)
+	logger.Fatal("failed to start payment server: %v\n",
+		pmt.ListenAndServe())
 }
--- a/payment/hash.go
+++ b/payment/hash.go
@@ -1,0 +1,32 @@
+package payment
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+// Computes hash of a payload.
+func hash(pld *payload) []byte {
+	data := crypto.Keccak256(
+		crypto.Keccak256(
+			[]byte("string message_id"),
+			[]byte("address receiver"),
+			[]byte("uint32 block_created"),
+			[]byte("bytes32 offering_hash"),
+			[]byte("uint192 balance"),
+			[]byte("address contract"),
+		),
+		crypto.Keccak256(
+			[]byte("Privatix: sender balance proof signature"),
+			[]byte(pld.AgentAddress),
+			big.NewInt(int64(pld.OpenBlockNumber)).Bytes(),
+			[]byte(pld.OfferingHash),
+			[]byte(pld.Balance),
+			[]byte(pld.ContractAddress),
+		),
+	)
+	msg := fmt.Sprintf("\x19Ethereum Signed Message:\n%d%s", len(data), data)
+	return crypto.Keccak256([]byte(msg))
+}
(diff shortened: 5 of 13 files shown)


dappctrl-test.config.json
```diff
@@ -25,5 +26,10 @@
         "CertFile": "dappctrl.cert",
         "KeyFile": "dappctrl.key",
         "TLS": false
+    },
+
+    "PaymentServer": {
+        "Addr": "localhost:9000",
+        "TLS": null
```
ababo: Why not `false`?

data/schema.go
```diff
@@ -15,6 +19,16 @@ type Subject struct {
...
+// PrivateKeyBytes returns private key's binary representation.
+func (s *Subject) PrivateKeyBytes() ([]byte, error) {
+	return base64.URLEncoding.DecodeString(*s.PrivateKey)
+}
+
+// PublicKeyBytes returns private key's binary representation.
+func (s *Subject) PublicKeyBytes() ([]byte, error) {
+	return base64.URLEncoding.DecodeString(s.PublicKey)
+}
+
```
ababo: I guess we shouldn't have separate functions for private and public case. I suggest adding a new `db/misc.go` file and putting there:
```go
func ToBytes(string data) ([]byte, error) {
// ...
}

func FromBytes([]byte data) string {
// ...
}
```

data/schema.go
```diff
@@ -82,6 +96,21 @@ type Channel struct {
 	ReceiptSignature string `reform:"receipt_signature"`
 }
 
+// IsOpen returns true if channel's has open state.
+func (ch *Channel) IsOpen() bool {
```
ababo: Really redundant. If it wasn't so, I was suggesting renaming to `Open()` to be more go-idiomatic.

data/schema.go
```diff
@@ -82,6 +96,21 @@ type Channel struct {
...
+	return ch.State == ChannelOpen
+}
+
+// TotalDepositNum returns total deposit as eth's number.Number.
+func (ch *Channel) TotalDepositNum() (*number.Number, error) {
```
ababo: Please consider not to have this method (also seems to be redundant, though not so obviously as the previous). Otherwise please move all schema type methods into `db/misc.go`.

data/schema.go
```diff
@@ -82,6 +96,21 @@ type Channel struct {
...
+	return util.Base64ToEthNum(ch.TotalDeposit)
+}
+
+// ReceiptBalanceNum returns total deposit as eth's number.Number.
+func (ch *Channel) ReceiptBalanceNum() (*number.Number, error) {
```
ababo: The same is here.

data/schema.sql
```diff
@@ -61,7 +61,7 @@ CREATE TABLE channels (
     agent uuid NOT NULL REFERENCES subjects(id),
     client uuid NOT NULL REFERENCES subjects(id),
     offering uuid NOT NULL REFERENCES offerings(id),
-    block int NOT NULL,
+    block bigint NOT NULL,
```
ababo: I guess there's no need: current block number is incremented every 15 seconds. I doubt it will overflow at any visible future.

data/test.go
```diff
@@ -24,3 +31,44 @@ func NewTestDB(conf *DBConfig, logger *util.Logger) *reform.DB {
...
+// NewTestSubject returns new subject
+func NewTestSubject() *Subject {
+	priv, _ := ecdsa.GenerateKey(crypto.S256(), cryptorand.Reader)
+	b := crypto.FromECDSA(priv)
+	privB64 := base64.URLEncoding.EncodeToString(b)
+	priv, _ = crypto.ToECDSA(b)
+	return &Subject{
+		ID:         util.NewUUID(),
+		PrivateKey: &privB64,
+		PublicKey: base64.URLEncoding.EncodeToString(
```
ababo: Suggest to prepare this value before the struct initialisation.

data/test.go
```diff
@@ -24,3 +31,44 @@ func NewTestDB(conf *DBConfig, logger *util.Logger) *reform.DB {
...
+}
+
+// NewTestOffering returns new offering.
+func NewTestOffering(agent *Subject) *Offering {
+	return &Offering{
+		ID:        util.NewUUID(),
+		Agent:     agent.ID,
+		Service:   ServiceVPN,
+		Supply:    1,
+		Signature: "",
```
ababo: This is the default field value, i.e. can be removed.

data/test.go
```diff
@@ -24,3 +31,44 @@ func NewTestDB(conf *DBConfig, logger *util.Logger) *reform.DB {
...
+func NewTestChannel(agent, client *Subject, offering *Offering,
+	balance, deposit int64, state string) *Channel {
+	return &Channel{
+		ID:             util.NewUUID(),
+		Agent:          agent.ID,
+		Client:         client.ID,
+		Offering:       offering.ID,
+		Block:          uint(rand.Int()),
+		State:          state,
+		TotalDeposit:   base64.URLEncoding.EncodeToString(number.Big(deposit).Bytes()),
```
ababo: As was stated we can have an utility function for byte conversions, so this expressions can be shortened.

main.go
```diff
@@ -2,18 +2,21 @@ package main
...
 type config struct {
-	DB         *data.DBConfig
-	Log        *util.LogConfig
-	VPNServer  *vpnsrv.Config
-	VPNMonitor *vpnmon.Config
+	DB            *data.DBConfig
+	Log           *util.LogConfig
+	VPNServer     *vpnsrv.Config
+	VPNMonitor    *vpnmon.Config
+	PaymentServer *payment.Config
```
ababo: Please honour the alphabetic field ordering (not very important, but would be nicer).

payment/hash.go
```diff
@@ -0,0 +1,32 @@
+package payment
+
+import (
+	"fmt"
+	"math/big"
+
+	"github.com/ethereum/go-ethereum/crypto"
```
ababo: Oh, it really makes sense to group imports. Good idea!

payment/hash.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// hash computes hash of a payload.
```
ababo: Please capitalise comments.

payment/middleware.go
```diff
@@ -0,0 +1,21 @@
+package payment
+
+import (
+	"net/http"
+)
+
+// http methods
```
ababo: Please capitalise. Do we really need this type? I would remove it.

payment/misc.go
```diff
@@ -0,0 +1,163 @@
+package payment
+
+import (
+	"encoding/base64"
+	"encoding/json"
+	"net/http"
+
+	"github.com/ethereum/go-ethereum/common/number"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/privatix/dappctrl/data"
```
ababo: Being inspired by your imports grouping I would suggest to have 3 groups:
1. Standard library imports.
2. Other packages imports.
3. Own packages imports.

This scheme is typically used for C++ includes and seems to be good.

payment/misc.go
```diff
@@ -0,0 +1,163 @@
...
+// serverError is a payment server error.
+type serverError struct {
+	// Code is a status code.
+	Code int `json:"code"`
+	// Message is a description of the error.
+	Message string `json:"message"`
+}
+
+var (
+	errInvalidPayload = &serverError{
```
ababo: Please consider exporting errors to give user ability to distinguish between them.

payment/misc.go
```diff
@@ -0,0 +1,163 @@
...
+		return false
+	}
+	return true
+}
+
+func (s *Server) verifySignature(w http.ResponseWriter,
+	ch *data.Channel, pld *payload) bool {
+
+	client := &data.Subject{ID: ch.Client}
+	if s.db.FindByPrimaryKeyTo(client, client.PKPointer()) != nil {
```
ababo: Why not `s.db.FindByPrimaryKeyTo(client, client.ID)`?

payment/misc.go
```diff
@@ -0,0 +1,163 @@
...
+	if !crypto.VerifySignature(pub, hash(pld), sig[:len(sig)-1]) {
+		s.replyError(w, errInvalidSignature)
+		return false
+	}
+	return true
+}
+
+func (s *Server) validateChannelForPayment(w http.ResponseWriter,
+	ch *data.Channel, pld *payload) bool {
+	return (s.validateChannelState(w, ch) &&
```
ababo: I would remove the parenthesis.

payment/pay_test.go
```diff
@@ -0,0 +1,168 @@
...
+		panic(err)
+	}
+	pld.BalanceMsgSig = base64.URLEncoding.EncodeToString(sig)
+	return pld
+}
+
+func sendTestRequest(pld *payload) *httptest.ResponseRecorder {
+	body := &bytes.Buffer{}
+	json.NewEncoder(body).Encode(pld)
+	r := httptest.NewRequest("POST", "/v1/pmtChannel/pay", body)
```
ababo: Suggest to have a constant for this URL path not to duplicate it.

dappctrl-test.config.json
furkhat: payment server config has pointer to `TLS` struct which has other tls required fields. I thought this might be better then a flag field. If you think flag is better I will change it to flag, because configs "similarity" is important. I'm open to discuss this.

payment/misc.go
furkhat: These errors doesn't have outside users, they used internally and get written to response.

payment/misc.go
furkhat: changed, thanks

payment/misc.go
furkhat: removed, thanks

payment/pay_test.go
furkhat: extracted to constant, thanks

data/schema.go
furkhat: ok, so we want to keep schema structs just as data transfer objects and not bind methods to them, is this what you are saying?

data/schema.go
furkhat: https://golang.org/src/go/types/predicates.go

Usually when reading name of a predicate function it sounds like a question. If I name it `Open` it will sound like an "action" which might be ambiguous.

data/schema.sql
furkhat: postgres int type max value is [`2147483647`](https://www.google.com/search?q=postgres+types+int+max+value). Contract uses [uint32](https://github.com/Privatix/dapp-smart-contract/blob/master/psc/contracts/PrivatixServiceContract.sol#L238) the max value of which is [`4294967295`](https://www.google.com/search?q=uint32+max+value). 

I changed it to `bigint` because `4294967295` > `2147483647` and it was the most suitable type from all ([posgres docs](https://www.postgresql.org/docs/10/static/datatype-numeric.html)).

data/test.go
furkhat: ok

main.go
furkhat: ok

data/test.go
furkhat: removed, thanks

payment/middleware.go
furkhat: removed, thanks.

payment/misc.go
furkhat: thats nice, agreed.

data/schema.sql
ababo: At the moment the current block number is `5229974`. So the number of seconds up to overflow is (2147483647 - 5229974) * 15 > 32 billions. This means that the overflow will occur in more than 1000 years. The period between neighbour blocks can become shorter with a time, but this won't make the overflow probability more real anyway.

data/schema.go
ababo: I just suggested to put them outside the `schema.go` and also said that we can have a single method for byte decoding instead of dedicated pair for private and public keys.

data/schema.go
ababo: OK, you convinced me.

dappctrl-test.config.json
ababo: I'm not clearly against your way, just not sure if we should have so nested config structure. Up to you, but please make it consistent (my code should handle it the same way as yours).

payment/misc.go
ababo: OK, sure you know it better.

PR #21, opened 2018-03-20
DB refactoring started. Still not production version.

sofabeat:


data/schema_0.6.sql
```diff
@@ -0,0 +1,156 @@
...
+    signature text NOT NULL, -- agent's signature
+    tpl_version int NOT NULL, -- template version
+    service_name varchar(64) NOT NULL, -- name of service
+    description text, -- description for UI
+    country char(2) NOT NULL, -- ISO 3166-1 alpha-2
+    supply int NOT NULL, -- maximum identical offerings for concurrent use through different state channels
+    unit_name varchar(10) NOT NULL, -- like megabytes, minutes, etc
+    unit_type unit_type NOT NULL, -- type of unit. Time or material.
+    billing_type bill_type NOT NULL, -- prepaid/postpaid
+    setup_price privatix_tokens, -- setup fee
```
furkhat: lets name it `setup_fee`?

data/schema_0.6.sql
```diff
@@ -0,0 +1,156 @@
...
+
+-- State channels.
+CREATE TABLE channels (
+    id uuid PRIMARY KEY,
+    agent uuid NOT NULL REFERENCES users(id),
+    client uuid NOT NULL REFERENCES users(id),
+    offering uuid NOT NULL REFERENCES offerings(id),
+    block int NOT NULL, -- block number, when state channel created
+    ch_status channel_status NOT NULL, -- status related to blockchain
+    service_status svc_status NOT NULL, -- operational status of service
```
furkhat: ~~ch_status~~ -> status

data/schema_0.6.sql
```diff
@@ -0,0 +1,156 @@
...
+
+-- Client sessions.
+CREATE TABLE sessions (
+    id uuid PRIMARY KEY,
+    channel uuid NOT NULL REFERENCES channels(id),
+    started timestamp with time zone NOT NULL, -- time, when session started
+    stopped timestamp with time zone, -- time, when session stopped
+    units_used bigint NOT NULL, -- total units used in this session.
+		seconds_consumed bigint NOT NULL, -- total seconds interval from started is recorded
+		last_used_time timestamp with time zone NOT NULL, -- time of last usage reported
```
furkhat: last_user ~~_time~~ timestamp with time zone NOT NULL, -- time of last usage reported

data/schema_0.6.sql
```diff
@@ -0,0 +1,156 @@
...
+-- Client sessions.
+CREATE TABLE sessions (
+    id uuid PRIMARY KEY,
+    channel uuid NOT NULL REFERENCES channels(id),
+    started timestamp with time zone NOT NULL, -- time, when session started
+    stopped timestamp with time zone, -- time, when session stopped
+    units_used bigint NOT NULL, -- total units used in this session.
+    seconds_consumed bigint NOT NULL, -- total seconds interval from started is recorded
+    last_used_time timestamp with time zone NOT NULL, -- time of last usage reported
+    server_ip inet,
```
ababo: Maybe to put this and below fields into `service_specific json` like in `offerings`?

furkhat: I think these field are to be filled automatically by us. To do so we need to refer them in our code, thus can't make part of generic json.

sofabeat: `units_used` and `seconds_consumed` are essential. One of them at a time. Other fields may be blank, but I believe that they will be in use in most cases.

data/schema_0.6.sql
sofabeat: There are two statuses: "blockchain" and "service life-cycle". So it seems to me good idea to clearly distinguish between them.

data/schema_0.6.sql
sofabeat: Firstly, I thought to accept your naming. But then I saw, feild `unit_price`, just below. I think it can be better to use "_price" suffix in both cases. It points you to idea, how overall pricing is formed and let you identify "PRIX" token fields more easily. If somebody, still not convinced, I'll rename it.

data/schema_0.6.sql
furkhat: I updated `ch_status` to `channel_status` and shortened type `channel_status` to `chan_status` so it looks consistent with `service_status svc_status`.

sofabeat:
In https://github.com/Privatix/dappctrl/pull/21/commits/b22f00db88aa75e836095ab10dd7dfda5fa0a632 I've removed transitional service statuses as well.

PR #26, opened 2018-03-27
DB scheme changes

sofabeat:
This version adds following objects: job queue, eth logs, eth transactions, endpoints messages.

### DIFF
--- a/data/schema_0.6.sql
+++ b/data/schema_0.6.sql
@@ -13,6 +13,9 @@ CREATE TYPE unit_type AS ENUM ('units','seconds');
 -- SHA3-256 in base64 (RFC-4648).
 CREATE DOMAIN sha3_256 AS char(44);
 
+-- Etehereum address
+CREATE DOMAIN eth_addr AS char(20);
+
  -- Ethereum's uint192 in base64 (RFC-4648).
 CREATE DOMAIN privatix_tokens AS char(32);
 
@@ -50,6 +53,30 @@ CREATE TYPE msg_status AS ENUM (
 	'msg_channel_published' -- published in messaging channel
 );
 
+-- Transaction statuses.
+CREATE TYPE tx_status AS ENUM (
+	'unsent', -- saved in DB, but not sent
+	'sent', -- sent w/o error to eth node
+	'mined', -- tx mined
+	'uncle' -- tx is went to uncle block
+);
+
+-- Job creator.
+CREATE TYPE job_creator AS ENUM (
+	'user', -- by user through UI
+	'billing_checker', -- by billing checker procedure
+	'bc_monitor', -- by blockchain monitor
+	'task' -- by another task
+);
+
+-- Job status.
+CREATE TYPE job_status AS ENUM (
+	'new', -- previously never executed
+	'failed', -- failed to sucessfully execute
+	'skipped', -- skipped by user
+	'done' -- successfully executed
+);
+
 -- Users are party in distributed trade.
 -- Each of them can play an agent role, a client role, or both of them.
 CREATE TABLE users (
@@ -73,9 +100,9 @@ CREATE TABLE templates (
 CREATE TABLE products (
     id uuid PRIMARY KEY,
     name varchar(64) NOT NULL,
-    offer_tpl_id uuid REFERENCES templates(id),
-    offer_auth_id uuid REFERENCES templates(id),
-    offer_access_id uuid REFERENCES templates(id),
+    offer_tpl_id uuid REFERENCES templates(id), -- enables product specific billing and actions support for Client
+    -- offer_auth_id uuid REFERENCES templates(id), -- currently not in use. for future use.
+    offer_access_id uuid REFERENCES templates(id), -- allows to identify endpoint message relation
     usage_rep_type usage_rep_type -- for billing logic. Reporter provides increment or total usage
 );
 
@@ -83,7 +110,7 @@ CREATE TABLE products (
 CREATE TABLE offerings (
     id uuid PRIMARY KEY,
     tpl uuid REFERENCES templates(id), -- corresponding template
-    product uuid NOT NULL REFERENCES products(id), -- enables product specific billing and actions support
+    product uuid NOT NULL REFERENCES products(id), -- enables product specific billing and actions support for Agent
     hash sha3_256 NOT NULL, -- offering hash
     status msg_status NOT NULL, -- message status
     agent uuid NOT NULL REFERENCES users(id),
@@ -149,4 +176,67 @@ CREATE TABLE contracts (
     type contract_type NOT NULL,
     version smallint, --version of contract. Greater means newer
     enabled boolean NOT NULL -- contract is in use
+);
+
+-- Endpoint messages. Messages that include info about service access.
+CREATE TABLE endpoints (
+    id uuid PRIMARY KEY,
+    tpl uuid REFERENCES templates(id), -- corresponding endpoint template
+    channel uuid NOT NULL REFERENCES channels(id), -- channel id that is being accessed
+    hash sha3_256 NOT NULL, -- message hash
+    status msg_status NOT NULL, -- message status
+    signature text NOT NULL, -- agent's signature
+    tpl_version int NOT NULL, -- template version
+    payment_receiver_address varchar(106), -- address ("hostname:port") of payment receiver. Can be dns or IP.
+    dns varchar(100),
+    ipv4 inet,
+    ipv6 inet,
+    username varchar(100),
+    password varchar(48),
+    nonce uuid NOT NULL, -- random number to get different hash, with same parameters
+    additional_params json -- all additional parameters stored as JSON
+);
+
+-- Job queue.
+CREATE TABLE jobs (
+    id uuid PRIMARY KEY,
+    task_name char(30) NOT NULL, -- name of task
+    status job_status NOT NULL, -- job status
+    parent_obj char(30) NOT NULL, -- name of object that relid point on (offering, channel, endpoint, etc.)
+    relid uuid NOT NULL, -- related object (offering, channel, endpoint, etc.)
+    created_at timestamp with time zone NOT NULL, -- timestamp, when job was created
+    not_before timestamp with time zone, -- timestamp, used to create deffered job
+    created_by job_creator NOT NULL, -- job creator
+    fail_count smallint, -- number of failures
+    try_count smallint -- number of times job was executed
+);
+
+-- Ethereum transactions.
+CREATE TABLE ethtxs (
+    id uuid PRIMARY KEY,
+    hash sha3_256 NOT NULL, -- transaction hash
+    method char(30) NOT NULL, -- contract method
+    status tx_status NOT NULL, -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
+    issued timestamp with time zone NOT NULL, -- timestamp, when tx was sent
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
+    addr_from eth_addr NOT NULL, -- from etehereum address
+    addr_to eth_addr NOT NULL, -- from etehereum address
+    nonce numeric, -- tx nonce field
+    gas_price bigint, -- tx gas_price field
+    gas bigint, -- tx gas field
+    tx_raw text -- raw tx as was sent
+);
+
+-- Ethereum events.
+CREATE TABLE ethlog (
+    id uuid PRIMARY KEY,
+    tx_hash sha3_256, -- transaction hash
+    status tx_status NOT NULL -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
+    block_number int, -- event block number field
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
+    addr eth_addr NOT NULL, -- address from which this log originated
+    data text, -- contains one or more 32 Bytes non-indexed arguments of the log
+    topics text -- array of 0 to 4 32 Bytes DATA of indexed log arguments.
 );
--- a/data/schema_0.6.sql
+++ b/data/schema_0.6.sql
@@ -14,7 +14,7 @@ CREATE TYPE unit_type AS ENUM ('units','seconds');
 CREATE DOMAIN sha3_256 AS char(44);
 
 -- Etehereum address
-CREATE DOMAIN eth_addr AS char(20);
+CREATE DOMAIN eth_addr AS char(28);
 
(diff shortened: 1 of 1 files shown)


data/schema_0.6.sql
```diff
@@ -13,6 +13,9 @@ CREATE TYPE unit_type AS ENUM ('units','seconds');
 -- SHA3-256 in base64 (RFC-4648).
 CREATE DOMAIN sha3_256 AS char(44);
 
+-- Etehereum address
+CREATE DOMAIN eth_addr AS char(20);
```
ababo: We encode binary data using base64, so it needs to have 28 bytes.

data/schema_0.6.sql
```diff
@@ -150,3 +177,66 @@ CREATE TABLE contracts (
...
+    id uuid PRIMARY KEY,
+    tpl uuid REFERENCES templates(id), -- corresponding endpoint template
+    channel uuid NOT NULL REFERENCES channels(id), -- channel id that is being accessed
+    hash sha3_256 NOT NULL, -- message hash
+    status msg_status NOT NULL, -- message status
+    signature text NOT NULL, -- agent's signature
+    tpl_version int NOT NULL, -- template version
+    payment_receiver_address varchar(106), -- address ("hostname:port") of payment receiver. Can be dns or IP.
+    dns varchar(100),
+    ipv4 inet,
```
ababo: Maybe just to have IP address string which can be written as IPv6 or IPv4?

data/schema_0.6.sql
```diff
@@ -150,3 +177,66 @@ CREATE TABLE contracts (
...
+    ipv6 inet,
+    username varchar(100),
+    password varchar(48),
+    nonce uuid NOT NULL, -- random number to get different hash, with same parameters
```
ababo: Why not to use UUID for that?

data/schema_0.6.sql
```diff
@@ -150,3 +177,66 @@ CREATE TABLE contracts (
...
+    additional_params json -- all additional parameters stored as JSON
+);
+
+-- Job queue.
+CREATE TABLE jobs (
+    id uuid PRIMARY KEY,
+    task_name char(30) NOT NULL, -- name of task
+    status job_status NOT NULL, -- job status
+    parent_obj char(30) NOT NULL, -- name of object that relid point on (offering, channel, endpoint, etc.)
+    relid uuid NOT NULL, -- related object (offering, channel, endpoint, etc.)
```
ababo: Maybe `rel_id` or `related_id`?

data/schema_0.6.sql
```diff
@@ -150,3 +177,66 @@ CREATE TABLE contracts (
...
+    created_at timestamp with time zone NOT NULL, -- timestamp, when job was created
+    not_before timestamp with time zone, -- timestamp, used to create deffered job
+    created_by job_creator NOT NULL, -- job creator
+    fail_count smallint, -- number of failures
+    try_count smallint -- number of times job was executed
+);
```
ababo: Didn't find something similar to `next_job_id`? How will job queue be built?

data/schema_0.6.sql
```diff
@@ -150,3 +177,66 @@ CREATE TABLE contracts (
...
+
+-- Ethereum transactions.
+CREATE TABLE ethtxs (
```
ababo: Maybe `eth_txs` or more explicit `eth_transactions`?

data/schema_0.6.sql
sofabeat: Will fix it. Thanks.

data/schema_0.6.sql
sofabeat: Ok. Will change it here and in docs.

data/schema_0.6.sql
sofabeat: Ok. Seems possible. Will remove this field.

ababo: I actually meant `id`.

data/schema_0.6.sql
sofabeat: Changing to `rel_id`.

data/schema_0.6.sql
sofabeat: Oh! It is good point. Firstly, it will run job in scope of object (`parent_obj` + `rel_id`). Then order is defined by `created_at` time stamp. FIFO.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    created_by job_creator NOT NULL, -- job creator
+    fail_count smallint, -- number of failures
+    try_count smallint -- number of times job was executed
+);
+
+-- Ethereum transactions.
+CREATE TABLE eth_txs (
+    id uuid PRIMARY KEY,
+    hash sha3_256 NOT NULL, -- transaction hash
+    method char(30) NOT NULL, -- contract method
```
HaySayCheese: 1. I'll suggest to change `char(30)` to `text`, because we can't (and even should not try) to predict contract methods names and their length.

2. It should allow null values. Not all transactions might be related to **our** contract.

data/schema_0.6.sql
```diff
@@ -50,40 +53,64 @@ CREATE TYPE msg_status AS ENUM (
 	'msg_channel_published' -- published in messaging channel
 );
 
+-- Transaction statuses.
+CREATE TYPE tx_status AS ENUM (
+	'unsent', -- saved in DB, but not sent
```
HaySayCheese: Is there any reason to add transaction into the DB before sending it?
How about the predicate "tx is in the table, so it was sent to the chain".

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    try_count smallint -- number of times job was executed
+);
+
+-- Ethereum transactions.
+CREATE TABLE eth_txs (
+    id uuid PRIMARY KEY,
+    hash sha3_256 NOT NULL, -- transaction hash
+    method char(30) NOT NULL, -- contract method
+    status tx_status NOT NULL, -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
```
HaySayCheese: I think, its a jobs responsibility to relate on the tx. 
Transaction itself should knows nothing about jobs related.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    issued timestamp with time zone NOT NULL, -- timestamp, when tx was sent
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
```
HaySayCheese: This value can be calculated at runtime: `block_number_confirmed = block_number + some_useful_period_for_us`.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    addr_from eth_addr NOT NULL, -- from etehereum address
+    addr_to eth_addr NOT NULL, -- from etehereum address
+    nonce numeric, -- tx nonce field
+    gas_price bigint, -- tx gas_price field
+    gas bigint, -- tx gas field
+    tx_raw text -- raw tx as was sent
```
HaySayCheese: `jsonb` instead of `text` can be more suitable here.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    hash sha3_256 NOT NULL, -- transaction hash
+    method char(30) NOT NULL, -- contract method
+    status tx_status NOT NULL, -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
+    issued timestamp with time zone NOT NULL, -- timestamp, when tx was sent
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
+    addr_from eth_addr NOT NULL, -- from etehereum address
+    addr_to eth_addr NOT NULL, -- from etehereum address
+    nonce numeric, -- tx nonce field
+    gas_price bigint, -- tx gas_price field
```
HaySayCheese: Does "gas price" can be fetched from tx info, returned by geth? 
As I know, it is separated, highly volatile parameter of the network itself. 
TX has "gas_used", and "cumulative_gas_used" parameters. 
But why wee need to store the gas price?

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    gas bigint, -- tx gas field
```
HaySayCheese: `gas` -> `gas_used` ? (to keep names equal to the ethereum itself, where is possible)

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    id uuid PRIMARY KEY,
+    hash sha3_256 NOT NULL, -- transaction hash
+    method char(30) NOT NULL, -- contract method
+    status tx_status NOT NULL, -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
+    issued timestamp with time zone NOT NULL, -- timestamp, when tx was sent
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
+    addr_from eth_addr NOT NULL, -- from etehereum address
+    addr_to eth_addr NOT NULL, -- from etehereum address
+    nonce numeric, -- tx nonce field
```
HaySayCheese: Why this field is needed for us? 
Is it related for our business logic?

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    gas bigint, -- tx gas field
+    tx_raw text -- raw tx as was sent
+);
+
+-- Ethereum events.
+CREATE TABLE eth_logs (
+    id uuid PRIMARY KEY,
+    tx_hash sha3_256, -- transaction hash
+    status tx_status NOT NULL -- tx status (custom)
+    job uuid REFERENCES jobs(id) -- corresponding endpoint template
```
HaySayCheese: Similarly to the TX table: I think, its a job responsibility to relate on the event. event itself should knows nothing about jobs related.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    block_number int, -- event block number field
+    block_number_confirmed int, -- block number, when tx considered confirmed. Always greater or equal to real block number.
```
HaySayCheese: The same as tx_block_number_confirmed

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    addr eth_addr NOT NULL, -- address from which this log originated
```
HaySayCheese: Why it is necessary for us, from which one address the event occurs? Should not we rely on the blockchain ideology "common chain on distributed nodes"?

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    addr_to eth_addr NOT NULL, -- from etehereum address
+    nonce numeric, -- tx nonce field
+    gas_price bigint, -- tx gas_price field
+    gas bigint, -- tx gas field
+    tx_raw text -- raw tx as was sent
+);
+
+-- Ethereum events.
+CREATE TABLE eth_logs (
+    id uuid PRIMARY KEY,
```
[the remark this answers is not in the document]
sofabeat: Discussed. Decided to leave as is to be same as other tables.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    tx_hash sha3_256, -- transaction hash
+    status tx_status NOT NULL -- tx status (custom)
```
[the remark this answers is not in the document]
sofabeat: Feel free to update status, as you need. If you need to have separate status enum for logs - do it.

data/schema_0.6.sql
```diff
@@ -150,3 +177,64 @@ CREATE TABLE contracts (
...
+    created_at timestamp with time zone NOT NULL, -- timestamp, when job was created
+    not_before timestamp with time zone, -- timestamp, used to create deffered job
+    created_by job_creator NOT NULL, -- job creator
+    fail_count smallint, -- number of failures
+    try_count smallint -- number of times job was executed
+);
+
+-- Ethereum transactions.
+CREATE TABLE eth_txs (
+    id uuid PRIMARY KEY,
```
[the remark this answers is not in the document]
sofabeat: Leaving it as is for similarity with other tables.

data/schema_0.6.sql
sofabeat: ok. Let's change it.

data/schema_0.6.sql
sofabeat: There are some edge cases, regarding:
1. failure during `tx` sending
2. possibility, we can use HD wallets (in future).
I personally prefer to store it, but agree that it is arguable. Let leave it till we get to the point of implementation and look one more on how to design it.

data/schema_0.6.sql
sofabeat: I think we talking about different parameters:
1. http://web3js.readthedocs.io/en/1.0/web3-eth.html#eth-sendtransaction
2. http://web3js.readthedocs.io/en/1.0/web3-eth.html#gettransactionreceipt
You talking about `2` - how many gas was used by this `tx`. I thought about `1` how many `gas` I want to spend and what `gas_price` I'm going to pay. I think we should do either:
1. add `gas_used` field to make some statistics (not sure we need it).
2. ignore it for a while
If you think, we should add it, let's discuss it in pm.

data/schema_0.6.sql
sofabeat: I've read in some cases, that it might be useful. So, I've planned a place for it :) E.g. if we would like to resend, transaction we should increase it.
One of a kind: https://medium.com/kinfoundation/making-sense-of-ethereum-nonce-sense-3858d5588c64
If, you think, we can ignore it for a while - ok. You may remove this field.

data/schema_0.6.sql
sofabeat: Discussed. Let's leave it. Blockchain monitor will always try to attach a task to event. In case of crash, it will try more, till there is no events left without `job`.

data/schema_0.6.sql
sofabeat: Correct. Status and block number should be changed. Please, do.

data/schema_0.6.sql
sofabeat: It is address of contract, not IP address of ethereum node. Feel free to change comment, if you want.

PR #29, opened 2018-03-29
schema migrated

furkhat:


data/schema.go
```diff
@@ -1,103 +1,219 @@
...
+// User is party in distributed trade.
 // It can play an agent role, a client role, or both of them.
-//reform:subjects
-type Subject struct {
+//reform:users
+type User struct {
 	ID         string  `reform:"id,pk"`
 	PublicKey  string  `reform:"public_key"`
 	PrivateKey *string `reform:"private_key"`
+	Kind       string  `reform:"kind"`
```
ababo: User can be both client and agent.

data/schema.go
```diff
@@ -1,103 +1,219 @@
...
+	TemplateAccess = "access"
+)
+
+// Template is a user defined structures.
+// It can be an offer, auth or access template.
+//reform:templates
+type Template struct {
+	ID   string `reform:"id,pk"`
+	Hash string `reform:"hash"`
+	Raw  []byte `reform:"raw"`
```
ababo: I think for binary data we agreed to use base64-encoded strings instead?

data/schema.go
```diff
@@ -1,103 +1,219 @@
...
 // Offering is a service offering.
 //reform:offerings
 type Offering struct {
-	ID        string `reform:"id,pk"`
-	Agent     string `reform:"agent"`
-	Service   string `reform:"service"`
-	Supply    uint   `reform:"supply"`
-	Signature string `reform:"signature"`
+	ID                 string  `reform:"id,pk"`
+	Template           *string `reform:"tpl"`     // Offering's.
```
ababo: Is it possible for offering not to have template?

data/schema.go
```diff
@@ -1,103 +1,219 @@
...
+	Template               string  `reform:"tpl"`
+	Channel                string  `reform:"channel"`
+	Hash                   string  `reform:"hash"`
+	Status                 string  `reform:"status"`
+	Signature              string  `reform:"signature"`
+	TemplateVersion        string  `reform:"tplVersion"`
+	PaymentReceiverAddress *string `reform:"payment_receiver_address"`
+	DNS                    *string `reform:"dns"`
+	IPAddress              *string `reform:"ip_addr"`
+	Username               *string `reform:"username"`
```
ababo: Why are username and password optional?

data/schema.go
furkhat: it is not binary data. It is json field.

ababo: Initially I chose to use base64-encoded strings for unification since first variants of the data schema used hashes as primary keys (reform does not support `[]byte` as PK). After we decided to use UUID as primary keys we can drop all base64 and to store raw `[]byte`. Anyway it should be consistent everywhere.

ababo: Why not string?

furkhat: for convenience.  `json` package works with []byte

data/schema.go
sofabeat: no

furkhat: ok, I will make it not null in sql.

ababo: thus we need `NOT NULL` and to remove the indirection.

data/schema.go
sofabeat: It is possible that external service will authenticate user, by it's public key only. I think it is sufficient to derive client public key from channel creation transaction to authenticate user.

data/schema.go
ababo: OK, regarding storing binary data let's postpone this decision (now base64-encoded as before).

data/schema.go
sofabeat: Yes. I think it is possible. Let's remove it for know and rethink. Thanks!

data/schema.go
sofabeat: Let's leave it this as is. We will discuss it separately.

PR #31, opened 2018-03-29
 Scheme improvements, constraints, eth specific scheme, settings table

HaySayCheese:


data/schema.sql
```diff
@@ -74,28 +80,33 @@ CREATE TYPE job_status AS ENUM (
...
+CREATE TABLE settings (
+    key text PRIMARY KEY,
+    value jsonb NOT NULL
+);
+
 -- Users are party in distributed trade.
 -- Each of them can play an agent role, a client role, or both of them.
 CREATE TABLE users (
-    id uuid PRIMARY KEY,
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
```
ababo: I would generate ids in controller (since the uniqueness is guaranteed). If you're still confident in this decision you can create a domain with `DEFAULT` to shorten stuff.

data/schema.sql
```diff
@@ -104,133 +115,167 @@ CREATE TABLE products (
...
     client uuid NOT NULL REFERENCES users(id),
     offering uuid NOT NULL REFERENCES offerings(id),
-    block int NOT NULL, -- block number, when state channel created
+    block INT NOT NULL -- block number, when state channel created
+        CONSTRAINT positive_block CHECK (channels.block > 0),
+
     channel_status chan_status NOT NULL, -- status related to blockchain
     service_status svc_status NOT NULL, -- operational status of service
-    service_changed_time timestamp with time zone, -- timestamp, when service status changed. Used in aging scenarios. Specifically in suspend -> terminating scenario.
+    service_changed_time TIMESTAMP WITH TIME ZONE, -- timestamp, when service status changed. Used in aging scenarios. Specifically in suspend -> terminating scenario.
```
ababo: In PostgreSQL documentation data types are not in upper case. I also think lowercase is better here it helps to distinguish between language constructs and (possibly user-defined) data types.

data/schema.sql
```diff
@@ -74,28 +80,33 @@ CREATE TYPE job_status AS ENUM (
...
+    in_use BOOLEAN DEFAULT TRUE -- this account is in use or not
 );
 
 -- Templates.
 CREATE TABLE templates (
-    id uuid PRIMARY KEY,
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
     hash sha3_256 NOT NULL,
-    raw json NOT NULL,
+    raw json NOT NULL, -- todo: [suggestion] use jsonb instead of json, it is faster and can be indexed
```
furkhat: jsonb seems to be cool, but for our use cases it is redundant complexity. We never query this json field, we will use it only to validate users input.

data/schema.sql
```diff
@@ -74,28 +80,33 @@ CREATE TYPE job_status AS ENUM (
...
     kind tpl_kind NOT NULL
 );
 
 -- Products. Used to store billing and action related settings.
 CREATE TABLE products (
-    id uuid PRIMARY KEY,
-    name varchar(64) NOT NULL,
+    id uuid PRIMARY KEY DEFAULT uuid_generate_v4(),
+    name VARCHAR(64) NOT NULL, -- todo: [suggestion] change type to "text"
```
furkhat: hm, why?

data/schema.sql
```diff
@@ -121,15 +127,29 @@ CREATE TABLE offerings (
...
     unit_price privatix_tokens NOT NULL,
-    min_units bigint NOT NULL, -- used to calculate min required deposit
-    max_unit bigint, -- optional. If specified automatic termination can be invoked
-    billing_interval int NOT NULL, -- every unit numbers, that should be paid, after free units consumed
-    max_billing_unit_lag int NOT NULL, --maximum tolerance for payment lag (in units)
-    max_suspended_time int NOT NULL, -- maximum time in suspend state, after which service will be terminated (in seconds)
-    max_inactive_time_sec bigint, -- maximum inactive time before channel will be closed
-    free_units smallint NOT NULL DEFAULT 0, -- free units (test, bonus)
+    min_units bigint NOT NULL -- used to calculate min required deposit
+        CONSTRAINT positive_min_units CHECK (offerings.min_units >= 0),
```
ababo: Why do we have explicit constraint names? Doesn't seem to be used.

data/schema.sql
```diff
@@ -171,7 +200,7 @@ CREATE TABLE contracts (
     address sha3_256 NOT NULL, -- ethereum address of contract
     type contract_type NOT NULL,
     version smallint, --version of contract. Greater means newer
-    enabled boolean NOT NULL -- contract is in use
+    enabled BOOLEAN NOT NULL -- contract is in use
```
ababo: So you still think types should be in upper case?

data/schema.sql
sofabeat: As far as I remember the only `bonus` you get, is more clear error SQL will return you, in case of failure.

data/schema.sql
sofabeat: No. @HaySayCheese, please, use lowercase.

PR #33, opened 2018-04-02
add suggestions

drew2a:
Summary:
* removed cognitive complexity check
* removed return statement count check
* increased method lines count

### DIFF
--- a/.codeclimate.yml
+++ b/.codeclimate.yml
@@ -18,7 +18,7 @@ checks:
       threshold: 250
 
   method-complexity:
-    enabled: true
+    enabled: false
     config:
       threshold: 5
 
@@ -30,7 +30,7 @@ checks:
   method-lines:
     enabled: true
     config:
-      threshold: 25
+      threshold: 40
 
   nested-control-flow:
     enabled: true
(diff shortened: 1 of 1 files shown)


.codeclimate.yml
```diff
@@ -30,15 +30,15 @@ checks:
   method-lines:
     enabled: true
     config:
-      threshold: 25
+      threshold: 30
```
HaySayCheese: I think it should be increased to 50-60 lines

.codeclimate.yml
```diff
@@ -30,15 +30,15 @@ checks:
...
 
   nested-control-flow:
     enabled: true
     config:
       threshold: 4
 
   return-statements:
-    enabled: true
+    enabled: false
```
HaySayCheese: Why it was disabled? IT may be useful thing.
The treshhold itself may be set to 5-6 for some corner cases.

.codeclimate.yml
drew2a: Are you sure? 50-60 lines for function means it may be very big and very complex.
In my editor function like this takes two screens.

.codeclimate.yml
drew2a: Ok. 
I just think this check is unuseful.

.codeclimate.yml
drew2a: ok, let's use 40

PR #34, opened 2018-04-03
Offerings table status and block number

sofabeat:
Added to table offerings:
1) Blockchain status, that respresents, if offering was deactivated/removed
2) Block number, that should be used to represent, when offering was last pushed up or to delete old offering sometime. Without it we don't know, how fresh it is.

### DIFF
--- a/data/schema.sql
+++ b/data/schema.sql
@@ -50,6 +50,13 @@ CREATE TYPE msg_status AS ENUM (
 	'msg_channel_published' -- published in messaging channel
 );
 
+-- Offering status
+CREATE TYPE offer_status AS ENUM (
+	'empty', -- saved in DB, but not published to blockchain
+	'register', -- in registration or registered in blockchain
+    'remove' -- being removed or already removed from blockchain
+);
+
 -- Transaction statuses.
 CREATE TYPE tx_status AS ENUM (
 	'unsent', -- saved in DB, but not sent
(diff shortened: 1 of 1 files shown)


data/schema.sql
```diff
@@ -115,9 +122,11 @@ CREATE TABLE offerings (
     product uuid NOT NULL REFERENCES products(id), -- enables product specific billing and actions support for Agent
     hash sha3_256 NOT NULL, -- offering hash
     status msg_status NOT NULL, -- message status
+    offer_status offer_status NOT NULL, -- offer status in blockchain
+    block_number_updated bigint
+        CONSTRAINT positive_block_number_updated CHECK (offerings.block_number_updated > 0), -- block number, when offering was updated
     agent uuid NOT NULL REFERENCES users(id),
     signature text NOT NULL, -- agent's signature
-    tpl_version int NOT NULL, -- template version
```
drew2a: @furkhat have removed this field in his branch.

data/schema.sql
```diff
@@ -211,7 +220,6 @@ CREATE TABLE endpoints (
     hash sha3_256 NOT NULL, -- message hash
     status msg_status NOT NULL, -- message status
     signature text NOT NULL, -- agent's signature
-    tpl_version int NOT NULL, -- template version -- todo: [suggestion] add constraint here. What is current versions range?
```
drew2a: and this

PR #39, opened 2018-04-06
Job queue with parallel processing.

ababo:
This change contains:
    1. Job-related DB schema improvements.
    2. Job queue type which supports addition and processing.
    3. Corresponding unit-tests.

    Resolves BV-147, BV-149.

### DIFF
--- a/data/schema.sql
+++ b/data/schema.sql
@@ -89,10 +89,18 @@ CREATE TYPE job_creator AS ENUM (
 -- Job status.
 DROP TYPE IF EXISTS job_status CASCADE;
 CREATE TYPE job_status AS ENUM (
-    'new', -- previously never executed
-    'failed', -- failed to successfully execute
-    'skipped', -- skipped by user
-    'done' -- successfully executed
+    'active', -- processing or to be processed
+    'done', -- successfully finished
+    'failed', -- failed: retry limit is reached
+    'canceled' -- canceled
+);
+
+-- Job related object types.
+DROP TYPE IF EXISTS related_type CASCADE;
+CREATE TYPE related_type AS ENUM (
+    'offering', -- service offering
+    'channel', -- state channel
+    'endpoint' -- service endpoint
 );
 
 DROP TABLE IF EXISTS settings CASCADE;
@@ -254,7 +262,7 @@ CREATE TABLE contracts (
 DROP TABLE IF EXISTS endpoints CASCADE;
 CREATE TABLE endpoints (
     id uuid PRIMARY KEY,
-    tpl uuid REFERENCES templates(id), -- corresponding endpoint template
+    template uuid NOT NULL REFERENCES templates(id), -- corresponding endpoint template
     channel uuid NOT NULL REFERENCES channels(id), -- channel id that is being accessed
     hash sha3_256 NOT NULL, -- message hash
     status msg_status NOT NULL, -- message status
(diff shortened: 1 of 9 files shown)


data/schema.sql
```diff
@@ -89,10 +89,55 @@ CREATE TYPE job_creator AS ENUM (
...
+    'channel', -- state channel
+    'endpoint' -- service endpoint
+);
+
+-- Job handler names.
+DROP TYPE IF EXISTS job_handler CASCADE;
+CREATE TYPE job_handler AS ENUM (
+    'preChannelCreate',
+    'afterChannelCreate',
+    'preChannelTopUp',
```
furkhat: @sofabeat Should we prefix them with `"Agent"/"Client"`?

job/queue.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+}
+
+func (q *Queue) processMain() error {
+	period := time.Duration(q.conf.CollectPeriod) * time.Millisecond
+
+	for {
+		if q.checkExit() {
+			return ErrQueueClosed
+		}
```
furkhat: Is this statement redundant?

job/queue.go
```diff
@@ -0,0 +1,293 @@
...
+func (q *Queue) processWorker(w workerIO) {
+	var err error
+	for err == nil {
+		job, ok := <-w.job
+		if !ok {
+			break
+		}
+
+		// Job was collected active, but delivered here with some delay,
+		// so make sure it's still relevant.
```
furkhat: This seems to be important. Did you write tests for this?

data/schema.sql
ababo: I just used names from the document.

job/queue.go
ababo: No, first one is needed for case when no active jobs are present, the second - otherwise.

job/queue.go
ababo: No specific test, but this case is covered in the existing tests.

data/schema.sql
sofabeat: 1. Yes. We should prefix. This will allow dappctrl not to mix Client's tasks with Agent's ones. Documentation fixed to include this change.
2. I personally, don't like idea of holding `job_handler names` as part of DB scheme. I think of software upgrade and need for 'ALTER database'. Maybe it is my personal bad experience with upgrades, which includes 'ALTER'.

ababo: 1. Will do that.
2. Make decision. This approach has both advantages and downsides. The first are keeping the handler names typed (not to spend tons of times looking for typos). The second you just mentioned.

ababo: I would prefer the current way, but I understand your concerns.

ababo:
@sofabeat @furkhat updated.

job/queue.go
```diff
@@ -0,0 +1,293 @@
...
+	return nil
+}
+
+// Add adds a new job to the job queue.
+func (q *Queue) Add(j *data.Job) error {
+	if err := q.checkDuplicated(j); err != nil {
+		return err
+	}
+
+	j.ID = util.NewUUID()
```
HaySayCheese: Why the ID is reset here?
`job` should already has an ID, no?

job/queue.go
```diff
@@ -0,0 +1,293 @@
...
+		err = q.db.Save(job)
+	}
+
+	if err != nil {
+		q.exit <- struct{}{}
+
+		// Make sure the main routine is not blocked passing a job.
+		select {
+		case <-w.job:
+		default:
```
HaySayCheese: There is no any report about failed job delivering, and no any retry.

data/schema.sql
sofabeat: 2. You're right. I thought about advantage, that you mentioned. I believe that this list will change in future several times, as we will add some features and jobs that will handle them as a result. 
a. So, I think it we should not keep them as `types` in DB. 
b. If not as `types`, where? 
Options:
1. code
2. file
3. DB
We talk a bit about that we will have different retry count and delay settings for each particular job type. It is not implemented, isn't it? So, we have here job handler structure that consists of:
- name
- max try_count
- try_period - period between retries
- first_start delay - delay on first try
We need to keep it somewhere. And this also can solve issue with mistyping the name. It can live in JSON file or inside DB as separate table `job_settings`. Keep settings in code, is not flexible, but quick and error prone. Keep settings in DB - possible, but we should query settings once on dapp start, to limit number of unnecessary reads. File - we also should read it once, but it is not convenient to keep settings in different places. So, I propose DB. What do you think guys?

job/queue.go
ababo: No need to duplicate this call in millions different places.

job/queue.go
ababo: This code is responsible for graceful worker exit. Yes, I differentiate between user and system errors. Missing job handler or database connection failure is considered as a serious hazard which triggers the job queue to stop processing. User can always restart the processing if needed (I think practically it won't help a lot).

data/schema.sql
ababo: Actually wouldn't be too afraid of typos, since we use the constants defined in the `data`-package.

ababo: @sofabeat Make your decision and I'll change it in a minute (very minor and simple change).

sofabeat: Let's remove handler name from DB. What about 
> settings for each particular job type

ababo: Currently I implemented optional individual settings (per job). If you think that should be per type we can do one of the following:
1. type settings from config go to handler descriptor table which will include name -> func mappings.
2. type settings in a dedicated DB table, if we need ability to update them in runtime.

IMO the first option is simpler and is more than enough.

ababo:
@sofabeat updated

PR #40, opened 2018-04-06
Feature/dc bv 157 billing

HaySayCheese:


drew2a:
@HaySayCheese take a look at https://codeclimate.com/github/Privatix/dappctrl/pull/40
9 issues were found by golint

drew2a:
About tests: if schema will change, it will be quite difficult to maintain them.

dappctrl-test.config.json
```diff
@@ -47,5 +47,9 @@
             "host": "localhost",
             "port": 5000
         }
+    },
+
+    "Billing": {
+        "SourceDataDir": "<full path to the project>/dappctrl/billing/tests/"
```
sofabeat: In readme.md of dappctrl project we have variable. Can it be used here as well? 
`psql -U postgres -d dappctrl -f $DAPPCTRL_DIR/data/schema.sql`

HaySayCheese:
@drew2a 

> take a look at https://codeclimate.com/github/Privatix/dappctrl/pull/40

There was only comments issues.

> About tests: if schema will change, it will be quite difficult to maintain them.

It would be really difficult, but there is no other way to test this logic other than on some test data. 
The main synchronization point between offering channels and billing is the DB, and it seems that there is no way to implement tests without need to populate it with some test data, otherwise - required conditions can't be checked.

PR #41, opened 2018-04-09
Feature/dc bv 179 eth contract fixes

HaySayCheese:


HaySayCheese:
@drew2a codeclimate seems to misreport code duplication in first 2 issues: there are different types of events created in code, specified by the codeclimate.

The rest 31 issues are about documenting publicly exported methods.
But the methods them selfs are named too obvious. For example: NewChannelCreatedEvent. Guess, what this method do? I think, them may be leaved without stupid doc strings, that only would repeat the methods names in different words order.

HaySayCheese:
@drew2a I need little help with licenses analyzer. Please, tell me, what's broken here?

HaySayCheese:
@ababo I has been deleted pre-generated contract files, and created a `go generate` step.

furkhat:
@HaySayCheese can't generate code on Mac, getting error:
```
eth/contract/gen.go:3: running "./tools/abigen": fork/exec ./tools/abigen: exec format error
```
I guess you built binary on linux and it's not going to work on os x.

drew2a:
@HaySayCheese don't worry about this license issues. 
FOSSA found two `GNU General Public License v3 (GPL-3)` and report about them to us.

drew2a:
About codeclimate. I have marked duplication issues as "Invalid".

Issues that reported by `golint`:
 `exported function should have a comment or be unexported`
should be fixed.

ababo:
> I guess you built binary on linux and it's not going to work on os x.

Do you use `abigen` binary built for Linux on Mac?

ababo:
It should be buildable and testable on Mac, this is the most common workspace for most of us.

HaySayCheese:
@ababo @furkhat 
I has been tested various ways to provide you binary build of `abigen` on your platform, for a while. But it seems, that this feature requires sudo privileges, because `abigen` is dependant on `go-dep`, that depends on `brew`, that is not default app on the mac. Also, there is a lock for the @furkhat on ethereum lib, so I decided to not to investigate time in it any more, and to provide solution that just works, may be for now. So I returned to the pre-built versions of the contracts in the repo. This is, at least, cross-platform solution, that doesn't brings any inconveniences to any one one of us. At the end, I removed ABI files from the repo. Actual version of them are always available in `dapp-exchange` repo. 

@sofabeat balance method was added, as you asked as quickly as I was able to implement it. 
JSON RPC implementation returns results in wei, so I decided to keep it in sync and to return results in wei too, not in ETH.

@furkhat I also added tests for `GetTransactionReceipt`, and little bit improved the response schema, as you need that. So now it should be clear about how to use it. Also I added one new method `GetTransactionByHash ` that returns little bit more info about transaction when it's considered as mined. Please, see this page for the details about their difference: https://github.com/ethereum/wiki/wiki/JSON-RPC At the end: it seems, that our test environment in some cases doesn't mines transactions as it should be (potentially because of a bug in ganache-cli), and transactions, that was spawned by contract methods calls, are not returned by the `GetTransactionReceipt` method. I spent about 2 hours in search for the workaround for this, but, at this moment, I doesn't know hot to beat it. I used contract creation transaction for the test purposes. I'll contact Timur for the advice how to deal with it.

@sofabeat during tests implementation I noticed that to be able to support Infura we need also the mechanism for checking public https certificates of their domain. I have not spent much time on it, but now it seems that it would be little bit tricky to connect current lib implementation to the Infura, because we do not support `https` at this moment, and Infura doesn't supports `http`.

PR #45, opened 2018-04-15
Auth endpoints 

furkhat:


agent/uisrv/product.go
```diff
@@ -10,15 +10,15 @@ import (
...
 	if r.Method == "PUT" {
-		s.handlePutProducts(w, r)
+		basicAuthMiddlewareFunc(s, s.handlePutProducts)(w, r)
 		return
 	}
 	if r.Method == "GET" {
-		s.handleGetProducts(w, r)
+		basicAuthMiddlewareFunc(s, s.handleGetProducts)(w, r)
 		return
 	}
```
ababo: The snippet above is duplicated many times. It can be shortened. Moreover it can be moved into a dedicated handler modify like you do with `ValidateMethod`.

agent/uisrv/server_test.go
```diff
@@ -16,6 +23,13 @@ var (
 )
 
 func TestMain(m *testing.M) {
+	// Disable middleware during tests.
+	basicAuthMiddlewareFunc = func(_ *Server, h http.HandlerFunc) http.HandlerFunc {
```
ababo: Why?

PR #50, opened 2018-04-18
Feature/dc bv 188 pkey encryption

HaySayCheese:
* Added support of hex PKey format.
* Added support of JSON pKey format.
* Added method for password update.
* Added ability to call contract methods by signing them from the keys storage directly.
* Added constraints for only one active account.

### DIFF
--- a/eth/keystore/accounts.go
+++ b/eth/keystore/accounts.go
@@ -1,0 +1,151 @@
+package keystore
+
+import (
+	"encoding/hex"
+	"errors"
+	"github.com/ethereum/go-ethereum/accounts"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/accounts/keystore"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"io/ioutil"
+	"math/big"
+	"os"
+)
+
+// AccountsManagerConf stores configuration, related to accounts manager.
+type AccountsManagerConf struct {
+
+	// Specifies path to the dir with encrypted private keystore of accounts.
+	KeystorePath string `json:"keystorePath"`
+}
+
+// Accounts manager provides safe access to the private key(s) of ethereum account.
+//
+// By default, current implementation enforces only one account at a time,
+// due to the current bossiness model, but there is a possibility of wider count of private keystore support.
+// (internally used keystore supports multiple accounts).
+type AccountsManager struct {
+	keystore *keystore.KeyStore
+}
+
+// NewAccountsManager returns account manager, configured with "conf".
+// It doesn't performs any checks of "conf" correctness.
+// In case of invalid "conf" - other methods would report corresponding error in runtime.
+func NewAccountsManager(conf *AccountsManagerConf) *AccountsManager {
+	return &AccountsManager{
+		keystore: keystore.NewKeyStore(
+			conf.KeystorePath,
+
+			// More details about used password encryption algorithm:
+			// https://github.com/Tarsnap/scrypt
+			// Current realisation uses "StandardScryptN" and "StandardScryptP"
+			// used in ethereum, by default.
+			keystore.StandardScryptN,
+			keystore.StandardScryptP),
+	}
+}
+
+// SetPrivateKeyFromHex creates account (private and public keystore pair) from private key in hex format.
+// Format example: "adfd7c50906423a1bc64f9927f5c05f2f8cdd815ffed78534ff2a1f15cbcae19"
+// This is the most used key format in ethereum, many wallets uses it as default format.
+func (m *AccountsManager) SetPrivateKeyFromHex(pKeyHex, passPhrase string) (accounts.Account, error) {
+	if len(m.keystore.Accounts()) != 0 {
+		return accounts.Account{}, errors.New(
+			"only one private key at a time is supported. " +
+				"please, remove already present one before initialising")
+	}
+
+	pKeyBytes, err := hex.DecodeString(pKeyHex)
+	if err != nil {
+		return accounts.Account{}, err
+	}
+
+	pKeyECDSA, err := crypto.ToECDSA(pKeyBytes)
+	if err != nil {
+		return accounts.Account{}, err
+	}
+
+	return m.keystore.ImportECDSA(pKeyECDSA, passPhrase)
+}
+
+// SetPrivateKeyFromJSON creates account (private and public keystore pair) from enc-json format.
+// used by geth by default. This format stores key and related constants in json file.
+// For the format details, please, see internally stored JSON file in keystore dir.
+func (m *AccountsManager) SetPrivateKeyFromJSON(filePath string, passPhrase string) (accounts.Account, error) {
+	if len(m.keystore.Accounts()) != 0 {
+		return accounts.Account{}, errors.New(
+			"only one private key at a time is supported. " +
+				"please, remove already present one before initialising")
+	}
+
+	jsonKey, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return accounts.Account{}, err
+	}
+
+	return m.keystore.Import(jsonKey, passPhrase, passPhrase)
+}
+
+// UpdatePassword encrypts account with new password.
+// Returns error in case if previous password is invalid or there are any other internal error.
+func (m *AccountsManager) UpdatePassword(previousPassPhrase, newPassPhrase string) error {
+	err := m.checkAccountPresence()
+	if err != nil {
+		return err
+	}
+
+	account := m.keystore.Accounts()[0]
+	return m.keystore.Update(account, previousPassPhrase, newPassPhrase)
+}
+
+// Transactor returns transaction signer object, that might be used for signing contract calls.
+func (m *AccountsManager) Transactor(passPhrase string) (*bind.TransactOpts, error) {
+	err := m.checkAccountPresence()
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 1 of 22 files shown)


HaySayCheese:
@drew2a i was unable to see them before PR creating. Fixed

drew2a:
@HaySayCheese nice. Thank you.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+type AccountsManagerConf struct {
+
+	// Specifies path to the dir with encrypted private keystore of accounts.
+	KeystorePath string `json:"keystorePath"`
+}
+
+// AccountsManager provides safe access to the private key(s) of ethereum account.
+//
+// By default, current implementation enforces only one account at a time,
+// due to the current bossiness model, but there is a possibility of wider count of private keystore support.
```
sofabeat: 1. Currently we aim to support `multiple accounts` at at time.
2. Typo `bossiness`
3. For alpha we may choose to have only single account support, but let's discuss it shortly.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+		return accounts.Account{}, err
+	}
+
+	return m.keystore.ImportECDSA(pKeyECDSA, passPhrase)
+}
+
+// SetPrivateKeyFromJSON creates account (private and public keystore pair) from enc-json format.
+// used by geth by default. This format stores key and related constants in json file.
+// For the format details, please, see internally stored JSON file in keystore dir.
+func (m *AccountsManager) SetPrivateKeyFromJSON(filePath string, passPhrase string) (accounts.Account, error) {
```
sofabeat: @lart5 @drew2a I see that @HaySayCheese has implemented **both methods** for new account creation:
-  from HEX
- from geth JSON backup file
1. I think we should support both methods of account creation.
2. Current function `SetPrivateKeyFromJSON()`uses file path as input parameter. 

I see two options: 
1. GUI will open file and pass JSON to back-end
2. GUI will pass file path to back-end (current way)

Please, suggest. I think that (1) is more robust. Note, that you already importing offering template. It will be simple code reuse I suppose.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
+package keystore
+
+import (
+	"encoding/hex"
```
ababo: Please separate standard, third-party and own packages with empty lines.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	"io/ioutil"
+	"math/big"
+	"os"
+)
+
+// AccountsManagerConf stores configuration, related to accounts manager.
+type AccountsManagerConf struct {
+
```
ababo: Do we need an empty line here?

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+//
+// By default, current implementation enforces only one account at a time,
+// due to the current bossiness model, but there is a possibility of wider count of private keystore support.
+// (internally used keystore supports multiple accounts).
+type AccountsManager struct {
+	keystore *keystore.KeyStore
+}
+
+// NewAccountsManager returns account manager, configured with "conf".
+// It doesn't performs any checks of "conf" correctness.
```
ababo: perform

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+// In case of invalid "conf" - other methods would report corresponding error in runtime.
```
ababo: I would replace 'would' with 'might'.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,153 @@
...
+	pKeyECDSA, err := crypto.ToECDSA(pKeyBytes)
+	if err != nil {
+		return accounts.Account{}, err
+	}
+
+	return m.keystore.ImportECDSA(pKeyECDSA, passPhrase)
+}
+
+// SetPrivateKeyFromJSON creates account (private and public keystore pair) from enc-json format.
+// used by geth by default. This format stores key and related constants in json file.
```
ababo: Used, JSON.

eth/keystore/accounts.go
HaySayCheese: > Current function SetPrivateKeyFromJSON()uses file path as input parameter.

@sofabeat there is also third option present: the method might accept `[]byte`, and so the back end would be able to transfer json directly. We even might create 2 separate methods from file-oriented processing and stream-oriented.

eth/utils/tests_common.go
```diff
@@ -0,0 +1,48 @@
...
+	"log"
+	"net/http"
+)
+
+// FetchPSCAddress is utility method for fetching contract address in test environment.
+func FetchPSCAddress() string {
+	truffleAPI := GethEthereumConfig().TruffleAPI
+	response, err := http.Get(truffleAPI.Interface() + "/getPSC")
+	if err != nil || response.StatusCode != 200 {
+		log.Fatal("Can't fetch PSC address. It seems that test environment is broken.")
```
ababo: Makes sense to log the error. Here and below.

eth/utils/tests_common.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	body, err := ioutil.ReadAll(response.Body)
+	defer response.Body.Close()
+	if err != nil {
+		log.Fatal("Can't read response body. It seems that test environment is broken.")
+	}
+
+	data := make(map[string]interface{})
+	json.Unmarshal(body, &data)
```
ababo: Error handling? Here and below.

eth/utils/tests_common.go
```diff
@@ -0,0 +1,48 @@
...
+
+	return data["contract"].(map[string]interface{})["address"].(string)
```
ababo: Better to simplify by splitting into two lines.

eth/utils/tests_common.go
```diff
@@ -0,0 +1,48 @@
...
+	"io/ioutil"
+	"log"
+	"net/http"
+)
+
+// FetchPSCAddress is utility method for fetching contract address in test environment.
+func FetchPSCAddress() string {
+	truffleAPI := GethEthereumConfig().TruffleAPI
+	response, err := http.Get(truffleAPI.Interface() + "/getPSC")
+	if err != nil || response.StatusCode != 200 {
```
ababo: Better to use `http.StatusOK`. Here and below.

eth/keystore/accounts.go
HaySayCheese: Why should I do this?

HaySayCheese: Why should I do this? 
`gofmt` would reorder the imports on the next run.

eth/utils/tests_common.go
HaySayCheese: is `log.Fatal` not enough?

eth/utils/tests_common.go
HaySayCheese: You are expecting so dramatic changes in HTTP specification?

eth/keystore/accounts.go
HaySayCheese: As you wish.

eth/keystore/accounts.go
HaySayCheese: As you wish.

eth/keystore/accounts.go
HaySayCheese: As you wish.

eth/keystore/accounts.go
HaySayCheese: As you wish.

eth/utils/tests_common.go
HaySayCheese: There might be only one case: 
test API for the ethereum environment doesn't responds as it was expected. Usually, it might be fixed by restarting the API itself.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,152 @@
...
+			"only one private key at a time is supported. " +
+				"please, remove already present one before initialising")
+	}
+
+	jsonKey, err := ioutil.ReadFile(filePath)
+	if err != nil {
+		return accounts.Account{}, err
+	}
+
+	return m.keystore.Import(jsonKey, passPhrase, passPhrase)
```
sofabeat: `m.keystore.Import(jsonKey, passPhrase, passPhrase)` should use separate passPhrases. Importer should provide old passphrase and new passphrase - is our system.password.

HaySayCheese: I was expecting that importer would provide it's current password, and that the pkey should be save under the same password, but OK, I'll fix it. It is simple.

eth/keystore/accounts.go
```diff
@@ -0,0 +1,152 @@
...
+			// used in ethereum, by default.
+			keystore.StandardScryptN,
+			keystore.StandardScryptP),
+	}
+}
+
+// SetPrivateKeyFromHex creates account (private and public keystore pair) from private key in hex format.
+// Format example: "adfd7c50906423a1bc64f9927f5c05f2f8cdd815ffed78534ff2a1f15cbcae19"
+// This is the most used key format in ethereum, many wallets uses it as default format.
+func (m *AccountsManager) SetPrivateKeyFromHex(pKeyHex, passPhrase string) (accounts.Account, error) {
```
sofabeat: Maybe we can use default ethereum function for this: https://github.com/ethereum/go-ethereum/blob/7c74e166b0ee14cf040374f522bc399841a11e69/crypto/crypto.go#L127

PR #52, opened 2018-04-24
account endpoints

furkhat:
Implements accounts endpoints.
Refactoring of `eth` package.
Contract addresses moved to config file. If `TruffleAPIURL` provided, contract addresses get updated from test truffle api.


eth/common_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	if response.Result.TransactionHash != transactionHash {
+		t.Fatal("Transaction hashes are not equal.")
+	}
+}
+
+func TestTransactionByHash(t *testing.T) {
```
ababo: There's a common abbreviation for transaction: `TX`.

PR #54, opened 2018-04-27
password encryption, storage. account endpoint updates

furkhat:
* Updated uisrv tests uses basic auth.
* Updated account create endpoint derives ethaddr from private key.
* Updated data.Account to encode base64 big int string to json in numeric form.
* uisrv get /settings do not return system settings.
* refactor pay package tests. Fix payment's channel selection.


agent/uisrv/generic.go
```diff
@@ -16,8 +16,9 @@ type queryParam struct {
 
 // getConf is a config for generic get handler.
 type getConf struct {
-	Params []queryParam
-	View   reform.View
+	Params    []queryParam
+	View      reform.View
+	Transform func(reform.Struct) interface{}
```
ababo: Why do we need that function?

data/pwdstorage.go
```diff
@@ -0,0 +1,30 @@
+package data
```
ababo: Maybe better to rename the file with `pwd_storage.go`?

data/schema.go
```diff
@@ -11,14 +11,14 @@ import (
...
-	PublicKey        string     `json:"publicKey" reform:"public_key"`
+	PublicKey        string     `json:"-" reform:"public_key"`
 	PrivateKey       string     `json:"-" reform:"private_key"`
 	IsDefault        bool       `json:"isDefault" reform:"is_default"`
 	InUse            bool       `json:"inUse" reform:"in_use"`
 	Name             string     `json:"name" reform:"name"`
 	PTCBalance       uint64     `json:"ptcBalance" reform:"ptc_balance"`
 	PSCBalance       uint64     `json:"psc_balance" reform:"psc_balance"`
-	EthBalance       string     `json:"ethBalance" reform:"eth_balance"`
+	EthBalance       B64BigInt  `json:"ethBalance" reform:"eth_balance"`
```
ababo: No sure if we should use used derived types for DB schema.

data/types.go
```diff
@@ -0,0 +1,20 @@
+package data
+
+import (
+	"fmt"
+	"math/big"
+)
+
+// B64BigInt is a base64 of big.Int that implements json.Marshaler.
+type B64BigInt string
```
ababo: Do we really need it? What are advantages?

furkhat: So there is eth balance field which can be a very big   positive number and stored as bytes encoded to base64. API Clients expect eth balance as numeric value.

Rather then writing transformation logic implementing `json.Marshaler` seemed to be the simplest way for me.

I am happy to hear alternative solution.

data/pwdstorage.go
furkhat: you are right, thxs

data/types.go
ababo: Seems to be elegant, just please make sure `reform` handles that properly.

PR #55, opened 2018-04-28
Feature/md bv 209

dzeckelev:
check pls

1. I finished the tests refactoring 
coverage: 65.9% of statements
2. added Dep
3. fix billing/monitor.go


billing/monitor_test.go
```diff
@@ -0,0 +1,461 @@
...
+	unitsUsed uint64, adjustTime int64, num int) (sessions []*data.Session) {
+	if num <= 0 {
+		return sessions
+	}
+
+	for i := 0; i <= num; i++ {
+		var curTime time.Time
+
+		if adjustTime != 0 {
+			curTime = time.Now().Add(time.Second * time.Duration(adjustTime))
```
ababo: The way below is simpler:

```go
curTime := time.Now()
if adjustTime != 0 {
        curTime = curTime.Add(time.Second * time.Duration(adjustTime))
}
```

billing/monitor_test.go
```diff
@@ -0,0 +1,461 @@
...
+	if _, err := NewMonitor(time.Second, nil, logger); err == nil {
+		panic("Monitor object with empty database is created")
+	}
+
+	mon, err := NewMonitor(time.Second, testDB, logger)
+	if err != nil {
+		panic(err)
+	}
+
+	if mon == nil {
```
ababo: If there wasn't any error the monitor object **must** be created.

billing/monitor_test.go
```diff
@@ -0,0 +1,461 @@
...
+// Second one has enough "total_deposit", that is greater than offering setup price.
+//
+// Expected result:
+// Channel 1 is selected for suspending.
+// Channel 2 is not affected.
+//
+// Description: this test checks first rule in HAVING block.
+func TestMonitor_VerifySecondsBasedChannels_TotalDeposit(t *testing.T) {
+	defer data.CleanTestDB(t, testDB)
+	fixture := newFixture(t)
```
ababo: Please consider to defer closing the fixture which will delete all its objects instead cleaning the database.

@furkhat btw we lost the ability to parallelise our tests because of this hairy approach.

billing/monitor_test.go
```diff
@@ -0,0 +1,461 @@
...
+// There are 2 active SECONDS-based channels.
+// First one has 3 sessions records, that used in total more seconds, than is provided by the offering.
+// Second one has 2 sessions records, that used less seconds, than provided by the offering.
+//
+// Expected result:
+// Channel 1 is selected for suspending.
+// Channel 2 is not affected.
+//
+// Description: this test checks second rule in HAVING block.
+func TestMonitor_VerifySecondsBasedChannels_MaxUnit(t *testing.T) {
```
ababo: Please follow the naming conventions.

billing/monitor_test.go
```diff
@@ -0,0 +1,569 @@
...
+		BillingTest *billingTestConfig
+	}
+	testDB  *reform.DB
+	testMon *Monitor
+)
+
+const testPassword = "test-password"
+
+type billingTestConfig struct {
+	Offer struct {
```
ababo: Please make separate structures.

billing/monitor_test.go
```diff
@@ -0,0 +1,569 @@
...
+	Channel struct {
+		SmallDeposit   uint64
+		MidDeposit     uint64
+		BigDeposit     uint64
+		EmptyBalance   uint64
+		EmptyUnitsUsed uint64
+	}
+}
+
+type testFixture struct {
```
ababo: Can you try to reuse the test fixture added in `BV-232` (in `data/test.go`)?

billing/monitor_test.go
```diff
@@ -0,0 +1,569 @@
...
+	template := data.NewTestTemplate(data.TemplateOffer)
+	data.InsertToTestDB(t, testDB, client, agent, product, template)
+	return &testFixture{
+		t:        t,
+		client:   client,
+		agent:    agent,
+		product:  product,
+		template: template,
+	}
+}
```
ababo: Please dilute this code with empty lines. Here and below.

billing/monitor_test.go
```diff
@@ -0,0 +1,569 @@
...
+
+	mon, err := NewMonitor(time.Second, testDB, logger)
+	if err != nil {
+		panic(err)
+	}
+	testMon = mon
+	os.Exit(m.Run())
+}
+
+func TestMonitor_VerifySecondsBasedChannels(t *testing.T) {
```
ababo: Please remove `_`.

billing/monitor_test.go
```diff
@@ -0,0 +1,569 @@
...
+	// than offering setup price.
+	// Second one has enough "total_deposit", that is greater
+	// than offering setup price.
+	//
+	// Expected result:
+	// Channel 1 is selected for suspending.
+	// Channel 2 is not affected.
+	//
+	// Description: this test checks first rule in HAVING block.
+	t.Run("LowTotalDeposit", func(t *testing.T) {
```
ababo: This needs refactoring. At least make these lambdas as separate functions. Test case (and any function as well) cannot be so huge and unreadable. The same for the stuff below.

billing/monitor_test.go
dzeckelev: 1) I do not want to create extra objects Offering and Channel
2) NewTestFixture creates less objects than I need
3)  For this reason, the function ```func (f *TestFixture) Close()``` does not delete all of my test objects. This negatively affects the tests

ababo: 1. I think two extra rows won't impact on performance, code simplification is more important here.
2./3. You can wrap the fixture with your code just to make it smaller, but this is not mandatory.

billing/monitor.go
```diff
@@ -234,9 +234,9 @@ func (m *Monitor) processEachChannel(query string, processor func(string) error)
 		rows.Scan(&channelUUID)
 		processor(channelUUID)
 
-		if m.testsSelectedChannelsIDs != nil {
-			m.testsSelectedChannelsIDs = append(m.testsSelectedChannelsIDs, channelUUID)
-		}
+		//if m.testsSelectedChannelsIDs != nil {
+		m.testsSelectedChannelsIDs = append(m.testsSelectedChannelsIDs, channelUUID)
```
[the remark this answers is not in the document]
dzeckelev: this function does not work correctly.
Steps:
1. Is the slice empty? If not empty, then initialize with an empty slice.
2. We get the values of rows
3. Add a value to the slice only if it is not empty. So we can not add the first value, because the slice is empty.
The same thing happens with all the other values.

billing/monitor_test.go
```diff
@@ -0,0 +1,626 @@
...
+
+func (f *testFixture) addTestObjects(testObjs []reform.Record) {
+	data.SaveToTestDB(f.t, testDB, testObjs...)
+
+	f.testObjs = testObjs
+}
+
+func (f *testFixture) clean() {
+	records := append([]reform.Record{}, f.client, f.agent,
+		f.product, f.template)
```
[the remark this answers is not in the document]
dzeckelev: I was guided by the comment above.
```
@ababo 
Please consider to defer closing the fixture which will delete all its objects instead cleaning the database.

@furkhat btw we lost the ability to parallelise our tests because of this hairy approach.
```
@furkhat I did not understand correctly?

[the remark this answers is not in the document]
furkhat: Oh, I missed that part. Thanks.

billing/monitor_test.go
dzeckelev: fix it

PR #57, opened 2018-04-30
Agent <--> Client message encryption.

furkhat:


ababo:
Please add some commit description (at least mention the issue it resolves to automate JIRA promotion).

crypto/crypto.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"crypto/ecdsa"
+	"crypto/rand"
+	"fmt"
+
+	ethcrypto "github.com/ethereum/go-ethereum/crypto"
+	"github.com/ethereum/go-ethereum/crypto/ecies"
+)
+
+var sigLen = 64
```
ababo: Maybe `const`?

PR #60, opened 2018-05-03
Feature/md bv 243

dzeckelev:
added `config` column for `product` table

### DIFF
--- a/data/schema.sql
+++ b/data/schema.sql
@@ -144,7 +144,8 @@ CREATE TABLE products (
     is_server boolean NOT NULL, -- product is defined as server (Agent) or client (Client)
     salt bigint NOT NULL, -- password salt
     password sha3_256 NOT NULL,
-    client_ident client_ident_type NOT NULL
+    client_ident client_ident_type NOT NULL,
+    additional_params json -- all additional parameters stored as JSON --
 );
 
 -- Service offerings.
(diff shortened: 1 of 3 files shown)


data/schema.sql
```diff
@@ -144,7 +144,8 @@ CREATE TABLE products (
     is_server boolean NOT NULL, -- product is defined as server (Agent) or client (Client)
     salt bigint NOT NULL, -- password salt
     password sha3_256 NOT NULL,
-    client_ident client_ident_type NOT NULL
+    client_ident client_ident_type NOT NULL,
+    config json  -- Store configuration of product --
```
furkhat: not null?

dzeckelev: @furkhat as I understand the configuration can be downloaded after creating the object in database. Should it not null?

furkhat: It is basically not null all the time, because postgres json considers null as invalid json. (At least this is what I've seen :)

dzeckelev: I was guided by the fact that `endpoints.additional_params` and `offerings.additional_params` do not have `NOT NULL`.
`json` fields do not participate in the index, in contrast to `jsonb`, but if you think what you need here I would not mind.
On the other hand, Null values are always not very good.
@ababo what is your opinion?

PR #63, opened 2018-05-05
added sesssrv endpoint for product config upload (w tests)

dzeckelev:


sesssrv/error.go
```diff
@@ -6,10 +6,12 @@ import (
...
 const (
-	ErrCodeChannelNotFound  = srv.ErrCodeMax + 1
-	ErrCodeBadAuthPassword  = iota
-	ErrCodeNonActiveChannel = iota
-	ErrCodeSessionNotFound  = iota
+	ErrCodeChannelNotFound            = srv.ErrCodeMax + 1
+	ErrCodeBadAuthPassword            = iota
+	ErrCodeNonActiveChannel           = iota
+	ErrCodeSessionNotFound            = iota
+	ErrCodeProductConfAlreadyUploaded = iota
```
ababo: Why can't we upload config multiple times?

sesssrv/server.go
```diff
@@ -30,10 +30,11 @@ type Server struct {
...
 const (
-	PathAuth   = "/session/auth"
-	PathStart  = "/session/start"
-	PathStop   = "/session/stop"
-	PathUpdate = "/session/update"
+	PathAuth          = "/session/auth"
+	PathStart         = "/session/start"
+	PathStop          = "/session/stop"
+	PathUpdate        = "/session/update"
+	PathProductConfig = "/session/product/config"
```
ababo: Please separate your path with an empty line. Should be `/product/config`.

sesssrv/server_test.go
```diff
@@ -34,13 +41,15 @@ var (
 		SessionServerTest *testConfig
 	}
 
-	db     *reform.DB
-	server *Server
+	db         *reform.DB
+	server     *Server
+	allMethods = []string{PathAuth, PathStart, PathStop, PathUpdate,
```
ababo: Please name it as `allPaths` and separate it by an empty line.

sesssrv/error.go
dzeckelev: If you load the configuration several times, this can cause the configuration to be out of sync between the agent and the client.
Should I fix this?

ababo: Please consult with @sofabeat.

sofabeat: Let's allow pushing config multiple times. All responsibilty for out of sync config is on Agent. But it can be desired in some cases to resync. After syncing we always setting flag on client to pevent automatic syncing each time.

sesssrv/server_test.go
```diff
@@ -226,6 +258,8 @@ func TestNormalFlow(t *testing.T) {
 
 	testAuthNormalFlow(fxt)
 
+	productConfigNormalFlow(fxt)
```
ababo: This should not be here since it has nothing to do with client session. Please rename `TestNormalFlow` with `TestNormalSessionFlow` and put your tests into a separate file. I suggest the following two files: `session_test.go` and `product_test.go`.

sesssrv/server.go
```diff
@@ -30,10 +30,11 @@ type Server struct {
...
 const (
-	PathAuth   = "/session/auth"
-	PathStart  = "/session/start"
-	PathStop   = "/session/stop"
-	PathUpdate = "/session/update"
+	PathAuth          = "/session/auth"
+	PathStart         = "/session/start"
+	PathStop          = "/session/stop"
+	PathUpdate        = "/session/update"
+	PathProductConfig = "/product/config"
```
ababo: Please separate product path(s) from session paths by an empty line.

sesssrv/error.go
```diff
@@ -30,4 +31,8 @@ var (
 		Code:    ErrCodeSessionNotFound,
 		Message: "session not found",
 	}
+	ErrProductConfNotValid = &srv.Error{
```
ababo: Better would be:
```golang
ErrInvalidProductConf = &srv.Error{
        Code:    ErrCodeInvalidProductConf,
        Message: "invalid product configuration",
}
```

sesssrv/misc.go
```diff
@@ -27,6 +27,16 @@ func (s *Server) findProduct(
 	return &prod, true
 }
 
+func (s *Server) updateProduct(
+	w http.ResponseWriter, prod *data.Product) bool {
+	if err := s.db.Update(prod); err != nil {
+		s.Logger().Error("failed to update product: %v", err)
```
ababo: failed to update product config

dzeckelev: The sense in the errors is that the product has not been updated in the database, it can happen at all through the fault of the config.

PR #64, opened 2018-05-05
Implement some client UI API

kvap:
The PR adds:
1. `GET /ui/client/offerings` endpoint for the client. Supports filtering by range of unit prices and lists of countries.
2. `GET /ui/accounts/{id}/pkey` for both the client and the agent. It exports the private key of a given account as is (keystrore JSON).
3. The support of creating account by importing the private key in keystore JSON format at the `POST /ui/accounts/` endpoint.


agent/uisrv/account.go
```diff
@@ -29,6 +42,33 @@ func (s *Server) handleAccounts(w http.ResponseWriter, r *http.Request) {
...
+			w.WriteHeader(http.StatusNotFound)
+		} else {
+			w.WriteHeader(http.StatusBadRequest)
+		}
+		return
+	}
+
+	privKeyJsonBytes, err := data.ToBytes(acc.PrivateKey)
+	if err != nil {
+		s.replyUnexpectedErr(w)
```
ababo: Didn't you miss `return`?

agent/uisrv/account.go
```diff
@@ -37,29 +77,49 @@ func (s *Server) handleGetAccounts(w http.ResponseWriter, r *http.Request) {
...
+			s.logger.Warn("could not decode private key: %v", err)
+			s.replyInvalidPayload(w)
+			return
+		}
+		privKey, err = crypto.ToECDSA(pkBytes)
+		if err != nil {
+			s.logger.Warn("could not make ecdsa priv key: %v", err)
+			s.replyInvalidPayload(w)
+			return
+		}
```
ababo: Optional hint: putting an additional `return` here you can remove `else` which will improve readability.

agent/uisrv/account_test.go
```diff
@@ -113,6 +116,24 @@ func getTestAccountPayload(testAcc *truffle.TestAccount) *accountCreatePayload {
...
+func getTestAccountKeyStorePayload(testAcc *truffle.TestAccount) *accountCreatePayload {
+	payload := &accountCreatePayload{}
+
+	pkB, _ := hex.DecodeString(testAcc.PrivateKey)
+	payload.JsonKeyStorePassword = "helloworld"
+	privKey, _ := crypto.ToECDSA(pkB)
+
+	pkB64, _ := data.EncryptedKey(privKey, payload.JsonKeyStorePassword)
+	jsonBytes, _ := data.ToBytes(pkB64)
+	payload.JsonKeyStoreRaw = string(jsonBytes)
```
ababo: Just for confidence: are you sure we can safely skip error handling in the lines above?

agent/uisrv/account_test.go
```diff
@@ -131,13 +152,31 @@ func testAccountFields(
 		t.Fatal("wrong in use stored")
 	}
 
-	createdPK, err := data.TestToPrivateKey(created.PrivateKey, testPassword)
-	if err != nil {
-		t.Fatal("failed to decrypt created account's private key: ", err)
-	}
+	if payload.PrivateKey != "" {
```
ababo: I would try generalize the production/testing cases. Lots of duplications.

agent/uisrv/generic.go
```diff
@@ -28,16 +29,49 @@ func (s *Server) handleGetResources(w http.ResponseWriter,
...
-			fmt.Sprintf("%s=%s",
-				param.Field,
-				s.db.Placeholder(len(eqs)+1)))
-		args = append(args, val)
+
+		if op == "in" {
+			subvals := strings.Split(val, ",")
+			for _, subval := range subvals {
+				args = append(args, subval)
+			}
```
ababo: `args = append(args, strings.Split(val, ",")...)`

agent/uisrv/generic.go
```diff
@@ -28,16 +29,49 @@ func (s *Server) handleGetResources(w http.ResponseWriter,
...
+			eqs = append(eqs,
+				fmt.Sprintf("%s %s (%s)",
+					param.Field,
+					op,
+					strings.Join(manyPlaceHolders(len(subvals)), ",")))
```
ababo: Please consider adding additional variables to reduce nesting.

agent/uisrv/account_test.go
kvap: Yes, if it fails to create a valid payload here, then the test that uses it will fail.

agent/uisrv/generic.go
kvap: This will not work because `[]interface{}` and `[]string` are not compatible.

ababo: Oh, right. Do we really need `args` as `[]interface{}` and not strings?

agent/uisrv/account.go
kvap: Fixed

agent/uisrv/generic.go
kvap: Actually, yes we need it as `[]interface` later to pass to `db.SelectAllFrom()`.

drew2a:
@kvap take a look at https://codeclimate.com/github/Privatix/dappctrl/pull/64

PR #56, opened 2018-04-29
Feature/md bv 214

dzeckelev:
- ParseLinesFromFile added for parsing any files (not very large)
- data/message pkg added

PR #65, opened 2018-05-07
Feature/md bv 233

dzeckelev:


ept/config.go
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	if c.Ping == "" {
+		c.Ping = defaultConfig.Ping
+	}
+
+	if c.ConnectRetry == "" {
+		c.ConnectRetry = defaultConfig.ConnectRetry
+	}
+}
```
ababo: I guess you don't need that. If you read config from file just fill the structure with default values and unmarshal as JSON.

ept/config.go
```diff
@@ -0,0 +1,169 @@
...
+
+func autogenFu() string {
+	return " # autogenerate option"
```
ababo: Why not a constant?

ept/config_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+const (
+	errGenConfig = "config file is empty"
+)
+
+func TestGetText(t *testing.T) {
+	out, err := testEMT.ParseConfig(samplesPath+
+		conf.EptTest.ConfValidCaValid, true)
+	if err != nil {
+		t.Fatal(err.Error())
```
ababo: Why don't you pass directly without `Error()`?

ept/ept.go
```diff
@@ -223,16 +146,34 @@ func (e *EndpointMessageTemplate) parseConfig(filePath string,
...
+	// function is thread-safe, mutex is not required
+	findCa := func() error {
+		// check ca key
+		ca := results[caNameFromConfig]
+		if ca == "" {
+			return ErrCertNotExist
+		}
+
+		// absolute path
+		absPath := filepath.Dir(filePath) +
```
ababo: Use `filepath.Join` instead of concatenation.

ept/ept_test.go
```diff
@@ -74,35 +76,40 @@ func TestParsingValidConfig(t *testing.T) {
...
+	_, err := testEMT.ParseConfig(samplesPath+
+		conf.EptTest.ConfValidCaInvalid, true)
 	if err == nil {
 		t.Fatal(errPars)
 	}
 }
 
 func TestInputFormat(t *testing.T) {
-	addParams, err := testEMT.ParseConfig(conf.EptTest.ConfValidCaValid)
+	addParams, err := testEMT.ParseConfig(samplesPath+
```
ababo: `filepath.Join()`

ept/ept_test.go
```diff
@@ -141,7 +148,8 @@ func TestInputFormat(t *testing.T) {
 }
 
 func TestGenerateCorrectMessage(t *testing.T) {
-	addParams, err := testEMT.ParseConfig(conf.EptTest.ConfValidCaValid)
+	addParams, err := testEMT.ParseConfig(samplesPath+
```
ababo: The same as above.

ept/misc.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"io/ioutil"
+	"strings"
+
+	"github.com/privatix/dappctrl/util"
+)
+
+// ValidNetworkAddress If string includes the address and port
+// and has a correct format (address:port), then returns true
+func ValidNetworkAddress(addr string) bool {
```
ababo: Better would be `ValidateNetAddr` or shorter `ValidateAddr`.

ept/templates/ovpn/client.go
```diff
@@ -0,0 +1,110 @@
+package ovpn
+
+// ClientConfig template for OpenVpn client configuration file
+const ClientConfig = `# tunX | tapX | null TUN/TAP virtual network device
```
ababo: Again you put configuration into go-files. Why?

dzeckelev: This case is different because it is not test data.
This template should not be changed manually by the user.
If this is taken out of the code, there will be an additional point of failure.
Alternatively, if the user incorrectly edits the field, the application will most likely give an error when parsing the wrong file.
Should I fix this?

ept/config.go
dzeckelev: Here there is no reading, but writing to a file.
Values are taken from the map. Their number may be insufficient to form the correct file. This code ensures that the corresponding values will be populated.

ept/templates/ovpn/client.go
ababo: Use [statik](https://github.com/rakyll/statik) or something similar. Don't forget to add a `go generate` step calling the utility.

ept/config.go
ababo: Then better to have default config a the same struct, to fill the default struct instance with the default values. I don't get why do we need those tedious duplicate statements.

dzeckelev: If part of the data is present in the map, and the ConnectRetry parameter is not present, then we need to substitute the default value only for the ConnectRetry parameter, and not change the whole configuration.

data/schema.go
```diff
@@ -305,3 +305,15 @@ type Job struct {
 	CreatedBy   string    `reform:"created_by"`
 	TryCount    uint8     `reform:"try_count"`
 }
+
+// EndpointMessageTemplate for a endpoint message check.
+//reform:endpoint_message_template
+type EndpointMessageTemplate struct {
```
dzeckelev: I will remove it from here.

data/test_data.sql
```diff
@@ -35,3 +35,16 @@ VALUES ('ae5deac9-44c3-4840-bdff-ca9de58c89f4', true, '', '',
     6012867121110302348, '7U9gC4AZsSZ9E8NabVkw8nHRlFCJe0o_Yh9qMlIaGAg=', 1, 1);
 
 END TRANSACTION;
+BEGIN TRANSACTION;
+
+CREATE TABLE endpoint_message_template (
```
dzeckelev: I will remove it from here.

dappctrl-test.config.json
```diff
@@ -43,21 +43,16 @@
...
-            "0x89f45a588f06d542b16020dd657486c7d163db03a2d599e4a606f8eb459e38b11",
-            "0x89f45a588f06d542b16020dd657486c7d163db03a2d599e4a606f8eb459e38g1"],
         "ValidHost": ["privatix.io:443", "127.0.0.1:443"],
-        "InvalidHost": ["privatix.io:44C", "priv#tix.io:443", "privatix.io%443"],
         "ConfValidCaValid": "conf.valid.ca.valid",
         "ConfInvalid": "conf.invalid",
         "ConfValidCaInvalid": "conf.valid.ca.invalid",
         "ConfValidCaEmpty": "conf.valid.ca.empty",
-        "ConfValidCaNotExist": "conf.valid.ca.notexist"
+        "ConfValidCaNotExist": "conf.valid.ca.notexist",
```
ababo: Please sort keys alphabetically.

ept/config/client.go
```diff
@@ -0,0 +1,158 @@
...
+	if err := json.Unmarshal(additionalParams, &params); err != nil {
+		return nil, err
+	}
+
+	config := defaultConfig
+
+	config.ServerAddress = srvAddr
+	config.Port = srvPort
+
+	for key, val := range params {
```
ababo: Why not to have a structure instead of this tedious initialisation?

ept/config/errors.go
```diff
@@ -0,0 +1,14 @@
+package config
+
+import "github.com/pkg/errors"
+
+// Config errors
+var (
+	ErrInput           = errors.New("one or more input parameters is wrong")
+	ErrFilePathIsEmpty = errors.New("filePath is empty")
+	ErrCertNotExist    = errors.New("certificate not exist in the config file")
```
ababo: doesn't exist

util/validation.go
```diff
@@ -69,6 +77,7 @@ var (
 //     - "ethHash" - Ethereum hash values
 //     - "ethAddr" - Ethereum address values
 //     - "netPort" - network port values
+//     - "tlsCert" - TLS certificate values
 func ValidateFormat(f Format, val string) error {
```
ababo: Don't see the reason why we need a dedicated function instead of more specialised ones. Adding new types will cause it to grow, so it will call format sub-functions under-the-hood. So this doesn't make any sense. Please replace if with ad-hoc functions.

ept/config/server.go
```diff
@@ -0,0 +1,219 @@
...
+func ServerConfig(filePath string, withCa bool,
+	keys []string) (map[string]string, error) {
+	if filePath == "" {
+		return nil, ErrFilePathIsEmpty
+	}
+	return parseConfig(filePath, keys, withCa)
+}
+
+func parseConfig(filePath string,
+	keys []string, withCa bool) (map[string]string, error) {
```
ababo: Please split this function, it's too long.

ept/config/server.go
```diff
@@ -0,0 +1,219 @@
...
+// PushConfig push OpenVpn config to Session Server
+// Hostname or ip address and port from Session Server
+// is taken from sessSrvConfig. Function can be canceled via context.
+// username and password needed for access the product in the database.
+// The confPath, caPath are absolute paths to OpenVpn config and Ca files,
+// In variable keys the list of keys for parsing.
+// The timeout in seconds between attempts to send data to the server must
+// be specified in variable retrySec
+func PushConfig(ctx context.Context, sessSrvConfig *srv.Config,
+	logger *util.Logger, username, password, confPath,
```
ababo: Makes sense to split this function too.

ept/config/server.go
```diff
@@ -0,0 +1,219 @@
...
+		select {
+		case <-ctx.Done():
+			return ErrCancel
+		case err := <-errC:
+			if err != nil {
+				logger.Warn("Failed to push app config to"+
+					" dappctrl. Original error: %s", err)
+				time.Sleep(time.Second *
+					time.Duration(retrySec))
+				continue
```
ababo: Why do we need this `continue`?

ept/config/server_test.go
```diff
@@ -0,0 +1,296 @@
...
+	if err := json.Unmarshal(prod.Config, &out); err != nil {
+		t.Fatal(err)
+	}
+
+	if !validParams(conf.EptTest.ExportConfigKeys, out) {
+		t.Fatal(errPars)
+	}
+}
+
+func TestPushConfigCancel(t *testing.T) {
```
ababo: These tree functions are almost identical. Please don't generate redundancies.

ept/service.go
```diff
@@ -0,0 +1,217 @@
...
+	case result := <-c:
+		return result.msg, result.err
+	case <-time.After(timeout):
+		close(done)
+		return nil, ErrTimeOut
+	}
+}
+
+func (s *Service) processing(req *req) {
+	resp := make(chan *result)
```
ababo: Split this function, it's huge and unreadable!

ept/service.go
```diff
@@ -0,0 +1,217 @@
...
+		if err := json.Unmarshal(prod.Config, &conf); err != nil {
+			select {
+			case <-req.done:
+				return
+			case resp <- newResult(nil, err):
+				return
+			}
+		}
+
+		if prod.OfferAccessID == nil {
```
ababo: I see lots of variants of the same code pattern, please generalize and abstract those.

ept/config/client.go
ababo: Why do you need a map?

dzeckelev: The `additionalParams` is the JSON object. The object deserializes in map.

util/validation.go
dzeckelev: @ababo
@sofabeat  proposed the idea of creating function to verify host addresses.  https://github.com/Privatix/dappctrl/pull/56/files/54a2b4c1f50a91adfc39725cb08602cc84407bf1#r184897261
I liked the idea of making a validator, but I added several more formats for validation.
Should I remove the validator?

ababo: Yes, please make separate functions instead.

ept/config/client.go
ababo: Is that mandatory? Since we will use only a predefined subset of the fields why not to keep it inside a dedicated structure? This will simplify this code. @HaySayCheese What do you think?

ept/config/server.go
dzeckelev: @ababo If you remove `continue` operator, the function will end in the next step. But this is not a valid flow, the function must perform the next iteration. 
it can be replaced
```
if err != nil {
    logger.Warn("Failed to push app config to dappctrl. Original error: %s", err)
    time.Sleep(time.Second * time.Duration(retrySec))
} else {
    return nil
}		
```

ept/config/server_test.go
dzeckelev: @ababo functions are different,
in one case the test server is started, in another case it does not start.
In the first case, the guarantee of delivery is checked.
In another case, the possibility of canceling a function through the context is checked.

ababo: I didn't say there're identical, I've said they're **almost** identical.

ept/service.go
dzeckelev: @ababo 
`OfferAccessID` a single field that can be zero.
Other fields will not cause panic when the next statement is executed
```
msg := Message{
			TemplateHash:           *prod.OfferAccessID,
			Username:               ch.ID,
			Password:               ch.Password,
			PaymentReceiverAddress: s.payAddr,
			ServiceEndpointAddress: "",
			AdditionalParams:       conf,
		}
```
the entire structure is validated below

ept/config/server_test.go
dzeckelev: @ababo I can make a wrapper to get rid of duplicate code

ababo: Do it please.

ept/service.go
ababo: Typically code duplications indicate some problems in decomposition.

ept/config/client.go
```diff
@@ -0,0 +1,154 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"io/ioutil"
+	"text/template"
+
+	"github.com/rakyll/statik/fs"
+
+	// it is necessary for work statik
```
ababo: Please this comment, or better just remove it.

ept/service.go
```diff
@@ -0,0 +1,299 @@
...
+			}
+		}
+
+		if exit {
+			return
+		}
+
+		go s.find(done, localErr, &prod, offer.Product)
+
+		select {
```
ababo: These three selects are suspiciously similar. Can you put such a select into a function and reuse this code?

ept/config/client.go
dzeckelev: this comment suggested writing codeclimate, it issues a warning that the comment is needed here

ababo: // This is necessary for statik.

ept/service.go
dzeckelev: function `s.find` already are wrapper, I can simplify using the pipeline pattern, but reading this code will be very difficult.

PR #67, opened 2018-05-09
Recover pub key from signature values & Contracts generation

furkhat:
* Delete some eth client methods, they are already implemented in go-ethereum. Fix agent/uisrv tests.
* Contracts generation.
* Recover pub key from signature values.


agent/uisrv/account_test.go
```diff
@@ -3,18 +3,20 @@
...
+	"context"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/common"
+
```
ababo: Please remove this empty line.

README.md
```diff
@@ -42,6 +42,7 @@ mkdir -p $DAPPCTRL_DIR
 git clone git@github.com:Privatix/dappctrl.git $DAPPCTRL_DIR
 go get -d $DAPPCTRL/...
 go get -u gopkg.in/reform.v1/reform
+go install github.com/ethereum/go-ethereum/cmd/abigen
```
ababo: Can we use `go get` here for unification?

agent/uisrv/server.go
```diff
@@ -3,9 +3,9 @@ package uisrv
 import (
 	"net/http"
 
+	"github.com/ethereum/go-ethereum/ethclient"
```
ababo: Please don't mix local and third-party imports.

eth/eth_test.go
```diff
@@ -3,16 +3,24 @@
...
+	"encoding/hex"
 	"os"
 	"testing"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/crypto"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/privatix/dappctrl/eth/contract"
```
ababo: Please separate local and third-party imports.

eth/util/util_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"bytes"
+	"os"
+	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+	apputil "github.com/privatix/dappctrl/util"
+)
```
ababo: The same as above.

agent/uisrv/server.go
```diff
@@ -47,12 +47,21 @@ type Server struct {
...
 	psc *contract.PrivatixServiceContract,
 	pwdStorage data.PWDGetSetter) *Server {
-	return &Server{conf, logger, db, ethClient, ptc, psc, pwdStorage,
-		data.EncryptedKey, data.ToPrivateKey}
+	return &Server{
+		conf,
+		logger,
+		db,
+		ethClient,
+		time.Second * 5,
```
kvap: The duration here is hardcoded. Maybe we should at least put it in a const and give it a name. It would be great to also move all of these parameters to `NewServer` inside `Config`, because it is a bit strange that we have the config parameter but almost nothing is tweaked through that.

kvap: It is fine as it is now though, the issue is not very related to the PR.

PR #78, opened 2018-05-23
Transfer OpenVpn config from Agent to Client

dzeckelev:
- Create OpenVPN ClientAfterEndpointMsgSOMCGet specific function, that prepares all Client OpenVPN config files
- Allow parsed OpenVPN config to be pushed to dappctrl automatically once.

Resolves BV-236
Resolves BV-244 (Finish)


ept/config/client.go
```diff
@@ -50,62 +52,119 @@ var defaultConfig = &CConf{
...
+	Cipher        string `json:"cipher"`
 	ServerAddress string
-	Port          string
-	PingRestart   string
-	Ping          string
-	ConnectRetry  string
-	Ca            string
-	CompLZO       string
+	Port          string `json:"port"`
+	PingRestart   string `json:"ping-restart"`
```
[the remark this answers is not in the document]
dzeckelev: This was done deliberately. Json field names correspond to the names in the OpenVpn configuration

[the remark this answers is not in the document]
ababo: Why do we need to preserve the original names? Don't see any sense.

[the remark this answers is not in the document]
dzeckelev: This simplifies the flow. We do not control the names of the fields that the server sends us.

svc/dappvpn/misc.go
```diff
@@ -66,3 +72,7 @@ func getCreds() (string, string) {
 
 	return user, pass
 }
+
+func writeConfig(name string, cfg *config) error {
```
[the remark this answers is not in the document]
dzeckelev: this function is needed here, it changes the `dappvpn.config.json` file

svc/dappvpn/misc.go
```diff
@@ -7,9 +7,15 @@ import (
...
 	"path/filepath"
+
+	"github.com/privatix/dappctrl/util"
 )
 
-const chanPerm = 0644
+const (
+	chanPerm = 0644
+
+	jsonIdent = "    "
```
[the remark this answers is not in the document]
dzeckelev: this is needed for `writeConfig` function, which should be here

ept/config/client.go
```diff
@@ -3,109 +3,164 @@ package config
...
-// ClientConfig returns config object
-func ClientConfig(srvAddr, srvPort string,
-	additionalParams []byte) (*CConf, error) {
-	if !isHost(srvAddr) || !util.IsNetPort(srvPort) {
-		return nil, ErrInput
+// ConfDeployer is needed to store information about the directory in which
+// the directories with configuration files are stored.
+type ConfDeployer struct {
+	rootDir string
+}
```
furkhat: I don't see usages of this type outside of `config` package. Let's not export types we don't need.

ept/config/client_test.go
```diff
@@ -47,18 +58,47 @@ func TestGetText(t *testing.T) {
...
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if isNotExist(filepath.Join(end, clientConfName)) ||
+			isNotExist(filepath.Join(end, clientAccessName)) {
+			t.Fatal(errDeployConfig)
+		}
+	}
 }
```
furkhat: How do you test the output of config generation?

ept/config/misc.go
```diff
@@ -104,3 +105,14 @@ func fillCa(results map[string]string, filePath string) error {
 
 	return nil
 }
+
+func createPath(target string) error {
+	return os.MkdirAll(target, pathPerm)
+}
+
+func isNotExist(object string) bool {
```
furkhat: `func notExist(location string) bool`

svc/dappvpn/pusher/pusher.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+
+	c "github.com/privatix/dappctrl/ept/config"
+	"github.com/privatix/dappctrl/util"
+	"github.com/privatix/dappctrl/util/srv"
+)
+
+// Config for pushing OpenVpn configuration.
+type Config struct {
+	ExportConfigKeys []string
```
furkhat: Please add comments describing the use of `ExportConfigKeys`

ept/config/client.go
dzeckelev: `rootDir`  does not change during the execution of the program.
Why not save `rootDir`  field once.
I see no reason to use global variables for this.

dappctrl-test.config.json
```diff
@@ -122,5 +126,11 @@
 
     "VPNMonitorTest": {
         "ServerStartupDelay": 10
+    },
+
+    "VPNMonitorPusherTest": {
```
ababo: I don't like "pusher" term in this context. What about "VPNConfigPusher" or "VPNConfigUploader"? Anyway you should use the "config" word in the name.

ept/config/client.go
```diff
@@ -3,109 +3,160 @@ package config
...
+		clientConfName)); err != nil {
+		return err
+	}
+
+	return ioutil.WriteFile(filepath.Join(dir, clientAccessName),
+		[]byte(fmt.Sprintf("%s\n%s\n", login, pass)),
+		filePerm)
+}
+
+func autogenFu() string {
```
ababo: What is that? Isn't it linked to the tests? If yes, what is a reason to have it here?

ept/config/client_test.go
```diff
@@ -2,36 +2,114 @@ package config
...
+	}
+
+	cliParams, err := parseConfig(confFile,
+		conf.EptTest.ExportConfigKeys, false)
+	if err != nil {
+		t.Fatal(err)
+	}
 
-	conf, err := ClientConfig(address[0], address[1], param)
+	if cfg.Cipher != cliParams[nameCipher] {
```
ababo: You still didn't get rid of these redundancies.

ept/config/server.go
```diff
@@ -106,7 +106,8 @@ func parseLine(keys map[string]bool,
 	str := strings.TrimSpace(line)
 
 	for key := range keys {
-		if !strings.HasPrefix(str, key) {
+		sStr := strings.Split(str, " ")
+		if sStr[0] != key {
```
ababo: Are you confident that sStr cannot be empty?

svc/dappvpn/main.go
```diff
@@ -141,6 +145,31 @@ func handleMonitor() {
...
+			if err := pusher.PushConfig(ctx, c); err != nil {
+				logger.Error("failed to send OpenVpn"+
+					" server configuration: %s\n", err)
+				return
+			}
+
+			conf.Pusher.Pushed = true
+
+			// rewrite dappvpn configuration file
+			if err := writeConfig(confFile, conf); err != nil {
```
ababo: Why do we store OpenVPN configuration in file?

ept/config/client.go
dzeckelev: this is not related to the tests.
https://golang.org/pkg/text/template/#FuncMap

ept/config/client_test.go
dzeckelev: this is a special test.
The code we discussed earlier is here https://github.com/Privatix/dappctrl/blob/1e3360dfab099897e7a84c4bff1402dde7320bc5/ept/config/client.go#L204

ept/config/server.go
dzeckelev: Yes it can not be empty.

svc/dappvpn/main.go
dzeckelev: We need to overwrite the configuration so that when I next start the program, do not parse the configuration again.

ababo: why not to parse it on every launch? I guess it isn't critical?

dzeckelev: Parsing files is quite a heavy function. Why do it every launch?

ababo: Simpler is better (unless it's a bottleneck).

ababo: By introducing a new config parameter you add additional complexity. As you wish.

PR #93, opened 2018-06-05
Implement REST endpoints handlers for Client UI

dzeckelev:
- Create UI back-end endpoints according to swagger specification.
- Implement REST endpoints handlers for Client UI

Resolves BV-296, BV-298.

### DIFF
--- a/dappctrl.prod.config.json
+++ b/dappctrl.prod.config.json
@@ -1,0 +1,65 @@
+{
+    "AgentServer": {
+        "Addr": "localhost:3000",
+        "TLS": null,
+        "EthCallTimeout": 5
+    },
+
+    "BlockMonitor": {
+        "CollectPause": 6,
+        "SchedulePause": 6,
+        "Timeout": 5
+    },
+
+    "DB": {
+        "Conn": {
+            "user": "postgres",
+            "dbname": "dappctrl",
+            "sslmode": "disable"
+        }
+    },
+
+    "Eth": {
+        "Contract" : {
+            "PTCAddrHex": "0xcA9a5951628486fAf8B9f58dB565E33ef9673394",
+            "PSCAddrHex": "0xa396e36ca7c36d7b74580f4ab6943704b8f2ee2e"
+        },
+        "GethURL": "https://rinkeby.infura.io/2B6KNL16rwlqe4fjM7BB"
+    },
+
+    "Job": {
+        "CollectJobs": 100,
+        "CollectPeriod": 1000,
+        "WorkerBufLen": 10,
+        "Workers": 0,
+        "TryLimit": 3,
+        "TryPeriod": 60000,
+        "Types": {
+            "clientPreChannelCreate": {
+                "TryLimit": 3,
+                "TryPeriod": 60000
+            }
+        }
+    },
+
+    "Log": {
+        "Level": "info"
+    },
+
+    "PayAddress": "http://localhost:9000/v1/pmtChannel/pay",
+
+    "PayServer": {
+        "Addr": "localhost:9000",
+        "TLS": null
+    },
+
+    "SessionServer": {
+        "Addr": "localhost:8000",
+        "TLS": null
+    },
+
+    "SOMC": {
+        "ReconnPeriod": 5000,
+        "URL": "ws://89.38.96.53:8080"
+    }
+}
--- a/uisrv/account.go
+++ b/uisrv/account.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"net/http"
 	"strings"
+	"time"
 
 	"github.com/ethereum/go-ethereum/accounts/keystore"
 	"github.com/ethereum/go-ethereum/crypto"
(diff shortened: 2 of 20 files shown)


uisrv/channel.go
```diff
@@ -42,16 +131,148 @@ func (s *Server) handleGetChannels(w http.ResponseWriter, r *http.Request) {
...
+func (s *Server) filter(conds []string) (constraints string) {
+	if len(conds) > 0 {
+		for k, v := range conds {
+			conds[k] = "channels." + v
+		}
+
+		constraints = " AND " + strings.Join(conds, " AND ")
+	}
+
+	return constraints
```
ababo: Weird decomposition: you prepend table prefixes and join condition parts in the same function. The function name is weird as well - you don't perform any filtering here.

uisrv/channel.go
```diff
@@ -42,16 +131,148 @@ func (s *Server) handleGetChannels(w http.ResponseWriter, r *http.Request) {
...
+func ethAddrFromBase64(addr string) string {
+	ethAddr, err := data.ToAddress(addr)
+	if err != nil {
+		ethAddr = common.Address{}
+	}
+	return ethAddr.String()
+}
+
+func formatTimeStr(tm *string) *string {
+	if tm != nil && *tm != "" {
```
ababo: Why not to have functions which take `string` by value instead?

uisrv/channel.go
```diff
@@ -42,16 +131,148 @@ func (s *Server) handleGetChannels(w http.ResponseWriter, r *http.Request) {
...
+		if err = rows.Scan(&item.ID, &item.Agent, &item.Client,
+			&item.Offering, &item.Deposit,
+			&item.ChStat.ServiceStatus, &item.ChStat.ChannelStatus,
+			&item.ChStat.LastChanged, &item.ChStat.MaxInactiveTime,
+			&item.Job.ID, &item.Job.Type, &item.Job.Status,
+			&item.Job.CreatedAt, &i.secUsage, &i.unitsUsage,
+			&item.Usage.MaxUsage, &i.unitType, &item.Usage.Unit,
+			&i.costSeconds, &i.costUnits); err != nil {
+			s.logger.Warn("failed to scan rows: %v", err)
+			s.replyUnexpectedErr(w)
```
ababo: Why not to directly load records using `reform` instead of fragile DTOs?

uisrv/channel.go
```diff
@@ -42,16 +131,148 @@ func (s *Server) handleGetChannels(w http.ResponseWriter, r *http.Request) {
...
+
+                       LEFT JOIN jobs job
+                       ON channels.id = job.related_id
+                 WHERE channels.agent NOT IN (SELECT eth_addr FROM accounts)
+                       AND channels.id = job.related_id
+		`
+	queryFooter := `
+		 GROUP BY channels.id, job.id, offer.setup_price, offer.unit_price,
+                       offer.unit_type, offer.unit_name, offer.max_inactive_time_sec
+		`
```
ababo: Do we need such a complexity? Looks scary. Maybe we can substitute that with simple channel selection?

uisrv/server.go
```diff
@@ -67,7 +67,7 @@ const (
 	authPath            = "/auth"
 	channelsPath        = "/channels/"
 	clientChannelsPath  = "/client/channels/"
-	clientOfferingsPath = "/client/offerings"
+	clientOfferingsPath = "/client/offerings/"
```
ababo: Some paths end with slash, other don't. This is inconsistent.

uisrv/channel.go
dzeckelev: In this context, such functions are not needed.

uisrv/channel.go
ababo: Obviously slower, but much less fragile. If the performance isn't a bottleneck I would prefer to use standard reform features since this way is easier to maintain and refactor.

uisrv/channel.go
dzeckelev: The function returns additional conditions for the query.
It applies a filter to the query.
Suggest your own version.

uisrv/channel.go
ababo: Yeah, probably this is intentional, just amazed a bit.

uisrv/channel.go
dzeckelev: Here is the bottleneck. The function works with the frontend and returns an array of objects.

ababo: No pagination? I guess it should be, otherwise it doesn't scale with all of your optimisations. I think we could return linked instances separately (array of involved offerings, channels, etc.) instead of joining which produces lots of duplicated values.

dzeckelev: I have what you are offering:
https://github.com/Privatix/dappctrl/blob/a681a45b163f5c52dfeeff23645a8d2ff3691923/uisrv/channel_test.go#L207
I use it in the test. now I want to know how slow it is

dappctrl.prod.config.json
```diff
@@ -0,0 +1,65 @@
+{
```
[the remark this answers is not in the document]
sofabeat: For simplicity and speed only. It is not perfect, but not bad to have such values in config for alpha.

uisrv/account.go
```diff
@@ -156,6 +157,21 @@ func (s *Server) handleCreateAccount(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if err := s.queue.Add(&data.Job{
+		ID:          util.NewUUID(),
+		Status:      data.JobActive,
```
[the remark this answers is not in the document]
sofabeat: In monitor I see same implementation. Will create task to refactor.

PR #97, opened 2018-06-08
Simplify endpoint message deployment.

dzeckelev:
Resolves BV-345, BV-346.

### DIFF
--- a/main.go
+++ b/main.go
@@ -11,6 +11,7 @@ import (
 	"github.com/privatix/dappctrl/eth/contract"
 	"github.com/privatix/dappctrl/execsrv"
 	"github.com/privatix/dappctrl/job"
+	vpncli "github.com/privatix/dappctrl/messages/ept/config"
 	"github.com/privatix/dappctrl/monitor"
 	"github.com/privatix/dappctrl/pay"
 	"github.com/privatix/dappctrl/proc"
@@ -43,6 +44,7 @@ type config struct {
 	Report        *bugsnag.Config
 	SessionServer *sesssrv.Config
 	SOMC          *somc.Config
+	VPNClient     *vpncli.Config
 }
 
 func newConfig() *config {
@@ -56,6 +58,7 @@ func newConfig() *config {
 		Report:        bugsnag.NewConfig(),
 		SessionServer: sesssrv.NewConfig(),
 		SOMC:          somc.NewConfig(),
+		VPNClient:     vpncli.NewConfig(),
 	}
 }
 
@@ -134,7 +137,8 @@ func main() {
 
 	handler, err := worker.NewWorker(logger, db, somcConn,
 		worker.NewEthBackend(psc, ptc, gethConn),
-		pscAddr, conf.PayAddress, pwdStorage, data.ToPrivateKey)
+		pscAddr, conf.PayAddress, pwdStorage,
+		data.ToPrivateKey, conf.VPNClient)
 	if err != nil {
 		panic(err)
 	}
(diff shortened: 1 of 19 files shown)


main.go
```diff
@@ -96,6 +99,8 @@ func main() {
 	reporter := bugsnag.NewClient(conf.Report, db, logger)
 	logger.Reporter(reporter)
 
+	vpncli.Dir = conf.VPNClient.ConfigDir
```
ababo: Please don't use global variables. Packages must be reentarable.

messages/ept/config/client.go
```diff
@@ -43,14 +36,25 @@ const (
...
-	PingRestart:   defaultPingRestart,
-	Port:          defaultServerPort,
-	Proto:         defaultProto,
-	ServerAddress: defaultServerAddress,
+var (
+	// Dir information about the location of the directory in which
+	// the directories with configuration files are stored.
+	Dir string
+
+	defaultConfig = &cConf{
```
ababo: Please use write `NewConfig()` which fills default values.

messages/ept/config/client.go
```diff
@@ -66,51 +70,40 @@ type cConf struct {
...
-		return "", ErrInput
-	}
-
-	switch r := record.(type) {
-	case *data.Channel:
-		if r.ID == "" {
-			return "", ErrInput
-		}
+// endpoint.Channel and endpoint.Password parameters.
+func Deploy(db *reform.DB, endpoint string) error {
```
ababo: Please rename with `DeployConfig`.

ababo: Please add an exported function which takes an endpoint ID and returns a config path. And reuse it in your code. Make sure `golint` doesn't produce warnings.

messages/ept/service_test.go
```diff
@@ -142,6 +144,13 @@ func TestValidEndpointMessage(t *testing.T) {
 	fxt := newFixture(t)
 	defer fxt.clean()
 
+	fxt.product.ServiceEndpointAddress = &strings.Split(
```
ababo: Please add a check for the split result (not to panic on empty slice).

main.go
dzeckelev: I need to change the implementation of ClientAfterEndpointMsgSOMCGet
Do you mind if I do this?

ababo: Why not to have a type with config instead of a plain function? You create it in `main.go` and pass the instance to the job worker.

dzeckelev: I meant it. What we see here is the result of our early arrangements. I delegate this functional to  the job worker.

messages/ept/config/client.go
```diff
@@ -66,51 +56,54 @@ type cConf struct {
...
+		PingRestart:   defaultPingRestart,
+		Port:          defaultServerPort,
+		Proto:         defaultProto,
+		ServerAddress: defaultServerAddress,
+	}
+}
+
+// NewConfig creates a default VPN client configuration.
+func NewConfig() *Config {
+	return &Config{""}
```
ababo: Maybe `/etc/openvpn/config` or something similar?

messages/ept/service_test.go
ababo: You still didn't add any check.

dzeckelev: I added this here
https://github.com/Privatix/dappctrl/blob/3c438a38bef5048aec327817cbd02277a8a997fb/messages/ept/service_test.go#L122

dzeckelev: I added a simple slice check

ababo: Please add some reasonable path default value in my comment above. I approve.

PR #99, opened 2018-06-10
Add "created at" timestamp to some jobs

sofabeat:


uisrv/account.go
```diff
@@ -213,6 +213,7 @@ func (s *Server) handleUpdateAccountBalance(w http.ResponseWriter, r *http.Reque
 
 	if err = s.queue.Add(&data.Job{
 		Type:        jobType,
+		CreatedAt:   time.Now(),
```
furkhat: May be it is better to set createdAt inside of `queue.Add`?

PR #100, opened 2018-06-11
Tasks for the client channel/endpoint setup flow.

ababo:
This change includes:

1. ClientPreChannelCreate implementation and test.
2. ClientAfterChannelCreate implementation and test.
3. ClientPreEndpointMsgSOMCGet implementation and test.
4. ClientAfterEndpointMsgSOMCGet implementation and test.
5. Miscellaneous utilities and refactoring bits.

Partially resolves BV-292.

### DIFF
--- a/proc/worker/client.go
+++ b/proc/worker/client.go
@@ -1,8 +1,292 @@
 package worker
 
-import "github.com/privatix/dappctrl/data"
+import (
+	"encoding/json"
+	"fmt"
+	"math/big"
 
-// ClientPreChannelCreate creates channel.
-func (w *Worker) ClientPreChannelCreate(job *data.Job) error {
+	"github.com/AlekSi/pointer"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/messages"
+	"github.com/privatix/dappctrl/messages/ept"
+	"github.com/privatix/dappctrl/statik"
+	"github.com/privatix/dappctrl/util"
+)
+
+func (w *Worker) clientPreChannelCreateCheckDeposit(
+	acc *data.Account, offer *data.Offering) (uint64, error) {
+	addr, err := data.ToAddress(acc.EthAddr)
+	if err != nil {
+		return 0, err
+	}
+
+	amount, err := w.ethBack.PSCBalanceOf(&bind.CallOpts{}, addr)
+	if err != nil {
+		return 0, err
+	}
+
+	deposit := offer.UnitPrice*offer.MinUnits + offer.SetupPrice
+	if amount.Uint64() < deposit {
+		return 0, ErrNotEnoughBalance
+	}
+
+	return deposit, nil
+}
+
+func (w *Worker) clientPreChannelCreateCheckSupply(
+	offer *data.Offering, offerHash common.Hash) error {
+	supply, err := w.ethBack.PSCOfferingSupply(&bind.CallOpts{}, offerHash)
+	if err != nil {
+		return err
+	}
+
+	if supply == 0 {
+		return ErrNoSupply
+	}
+
 	return nil
+}
+
+func (w *Worker) clientPreChannelCreateSaveTX(
+	job *data.Job, acc *data.Account, offer *data.Offering,
+	offerHash common.Hash, deposit uint64) error {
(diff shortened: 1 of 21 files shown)


proc/worker/client.go
```diff
@@ -1,8 +1,282 @@
...
+		return 0, err
+	}
+
+	amount, err := w.ethBack.PSCBalanceOf(&bind.CallOpts{}, addr)
+	if err != nil {
+		return 0, err
+	}
+
+	deposit := offer.UnitPrice*offer.MinUnits + offer.SetupPrice
+	if amount.Uint64() < deposit {
```
dzeckelev: it is better to convert the deposit to bigInt, this will exclude the border state when the amount is larger than Uint64

ababo: @sofabeat Is it practically possible?

proc/worker/client.go
```diff
@@ -1,8 +1,282 @@
...
+	if err := data.Insert(w.db.Querier, &ch); err != nil {
+		return err
+	}
+
+	agentAddr, err := data.ToAddress(offer.Agent)
+	if err != nil {
+		return err
+	}
+
+	tx, err := w.ethBack.PSCCreateChannel(&bind.TransactOpts{},
```
dzeckelev: why bind.TransactOpts empty? So it should be?

proc/worker/client.go
```diff
@@ -1,8 +1,282 @@
...
+		return nil, err
+	}
+
+	if !util.ValidateJSON(schema, mdata) {
+		return nil, fmt.Errorf(
+			"failed to validate endpoint for chan %s", ch.ID)
+	}
+
+	var msg ept.Message
+	json.Unmarshal(mdata, &msg)
```
dzeckelev: forgot to check error

ababo: Nope, I didn't. The message is validated before.

proc/worker/client.go
ababo: Not being really familiar with this stuff I just followed the rest of the similar code which passes default options to the smart contract calls. If you have concrete suggestions to put here, I will modify that.

proc/worker/client_test.go
```diff
@@ -1,29 +1,211 @@
...
+	defer env.close()
+
+	fxt := env.newTestFixture(t,
+		data.JobClientPreEndpointMsgSOMCGet, data.JobChannel)
+	defer fxt.Close()
+
+	swapAgentWithClient(t, fxt)
+
+	msg := ept.Message{
+		TemplateHash:           "test-hash",
```
dzeckelev: It may be better to make these fields in constant?

ababo: No, it probably won't, since it's not reused.

proc/worker/client.go
dzeckelev: add From and Context , it would be enough.

proc/worker/client.go
drew2a: Total supply of PRIX is `1275455`.
As I know we store PRIX as an integer and supports 8 decimals.
So, max integer will be `1275455*100000000 `

proc/worker/client.go
```diff
@@ -1,8 +1,282 @@
...
+		return err
+	}
+
+	tx, err := w.ethBack.PSCCreateChannel(&bind.TransactOpts{},
+		agentAddr, offerHash, big.NewInt(int64(deposit)))
+	if err != nil {
+		return err
+	}
+
+	return w.saveEthTX(job, tx, "CreateChannel",
```
dzeckelev: if this function returns an error, what will happen to the channel that you saved in the database above?

ababo: Good point! Updated.

proc/worker/client.go
furkhat: Use `bind.NewKeyedTransactor(agentPrivKet)`, where `agentPrivKey` is `*ecdsa.PrivateKey`

proc/worker/client.go
```diff
@@ -1,8 +1,286 @@
...
+
+func (w *Worker) clientPreChannelCreateSaveTX(
+	job *data.Job, acc *data.Account, offer *data.Offering,
+	offerHash common.Hash, deposit uint64) error {
+	agentAddr, err := data.ToAddress(offer.Agent)
+	if err != nil {
+		return err
+	}
+
+	tx, err := w.ethBack.PSCCreateChannel(&bind.TransactOpts{},
```
furkhat: Add `// TODO: set gas limit from conf`

PR #106, opened 2018-06-14
Add job ClientPreUncooperativeClose.

dzeckelev:
Resolves BV-363.

### DIFF
--- a/proc/worker/client.go
+++ b/proc/worker/client.go
@@ -1,13 +1,16 @@
 package worker
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/AlekSi/pointer"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
 	"gopkg.in/reform.v1"
 
 	"github.com/privatix/dappctrl/data"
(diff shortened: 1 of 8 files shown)


proc/worker/client.go
```diff
@@ -1,15 +1,18 @@
...
 	"fmt"
 	"math/big"
+	"time"
 
 	"github.com/AlekSi/pointer"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"gopkg.in/reform.v1"
 
+	"github.com/ethereum/go-ethereum/core/types"
```
furkhat: move this upper to the second block of imports

proc/worker/client_test.go
```diff
@@ -233,10 +233,47 @@ func TestClientAfterUncooperativeCloseRequest(t *testing.T) {
...
+	var ch data.Channel
+	data.FindInTestDB(t, db, &ch, "id", fxt.Channel.ID)
+	if ch.ChannelStatus != data.ChannelWaitUncoop {
+		t.Fatal("bad channel status")
+	}
+
+	var tx data.EthTx
+
+	data.FindInTestDB(t, db, &tx, "related_id", fxt.Channel.ID)
+	defer data.DeleteFromTestDB(t, db, &tx)
```
furkhat: you could use `defer env.deleteEthTx(t, fxt.job.ID)`

PR #128, opened 2018-06-21
Separate job scheduling in blockchain monitor according to user role in settings.

dzeckelev:
Resolves BV-407.

### DIFF
--- a/monitor/schedule.go
+++ b/monitor/schedule.go
@@ -82,6 +82,12 @@ func (m *Monitor) schedule(ctx context.Context, timeout int64,
 		return
 	}
 
+	agent, err := data.IsAgent(m.db.Querier)
+	if err != nil {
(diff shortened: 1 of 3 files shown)


monitor/schedule.go
```diff
@@ -133,6 +145,28 @@ func isOfferingRelated(el *data.EthLog) bool {
 	return len(el.Topics) > 0 && offeringRelatedEventsMap[el.Topics[0]]
 }
 
+func (m *Monitor) isAgent() (bool, error) {
```
sofabeat: It is better to make it in comon package like data, as it like will be used in some other cases, where user role check is required. Like: https://github.com/Privatix/dappctrl/blob/133fc73edcbbf0796f8ad6fc0e43c330295c4d32/data/misc.go#L75

PR #129, opened 2018-06-21
Add dappctr config option to write logs to file.

dzeckelev:
Resolves BV-402.

### DIFF
--- a/dappctrl-test.config.json
+++ b/dappctrl-test.config.json
@@ -148,7 +148,8 @@
     },
 
     "Log": {
-        "Level": "info"
+        "Level": "info",
+        "LogPath": ""
     },
 
     "PayServer": {
--- a/util/log.go
+++ b/util/log.go
@@ -3,23 +3,30 @@ package util
 import (
 	"errors"
 	gofmt "fmt"
+	"io"
 	"log"
 	"os"
+	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/privatix/dappctrl/report"
 )
+
+const logFilePerm = 0644
 
 // Logger to log internal events.
 type Logger struct {
 	logger *log.Logger
 	level  int
 	rep    report.Reporter
+	out    io.Writer
 }
 
 // LogConfig is a logger configuration.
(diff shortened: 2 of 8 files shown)


dappctrl-test.config.json
```diff
@@ -148,7 +148,8 @@
     },
 
     "Log": {
-        "Level": "info"
+        "Level": "info",
+        "Logpath": ""
```
ababo: "LogPath"

util/log.go
```diff
@@ -57,17 +64,55 @@ func parseLogLevel(lvl string) int {
...
+func createLogFile(path string) (*os.File, error) {
+	info, err := os.Stat(path)
+	if err != nil {
+		return nil, err
+	}
+	if !info.IsDir() {
+		return nil, err
+	}
+
+	fileName := gofmt.Sprintf("dappctrl-%d.log", time.Now().Unix())
```
ababo: I believe it's better to make more explicit timestamps as part of the names, e.g. `dappctrl-2018-06-21.log`. Much easier to use.

util/log.go
```diff
@@ -57,17 +64,55 @@ func parseLogLevel(lvl string) int {
...
 	}
 
-	return &Logger{
-		logger: log.New(os.Stderr, "", log.LstdFlags),
-		level:  lvl,
-	}, nil
+	logger := &Logger{level: lvl}
+
+	if conf.Logpath != "" {
+		file, err := createLogFile(conf.Logpath)
```
ababo: Why do we need to save file instead or redirecting STDERR? You don't implement rotation, so this is almost equivalent.

dzeckelev: This will be solved in another task, when a log rotation will be implemented.

ababo:
If we're going to make a full-fledged logger with file rotation I suggest to choose an existing logger to work under the `util.Logger` hood. Indeed it's not so trivial to implement such rotation properly. Otherwise I don't see any sense to support explicit file output while having output to STDERR which can be easily redirected.

PR #138, opened 2018-06-23
sync method to get endpoint

furkhat:


ababo:
@furkhat Why did you merge without addressing my change request?

PR #139, opened 2018-06-23
update balances jobs recreates itself instead of failure

furkhat:


proc/worker/common.go
```diff
@@ -185,8 +186,9 @@ func (w *Worker) AccountAddCheckBalance(job *data.Job) error {
 	}
 
 	// HACK: return error to repeat job after a minute.
+	job.ID = util.NewUUID()
 	job.NotBefore = time.Now().Add(time.Minute)
-	return fmt.Errorf("repeating job")
```
ababo: > job.NotBefore = time.Now().Add(time.Minute)

I don't think this value should be hardcoded.

PR #146, opened 2018-06-26
ClientPreUncooperativeClose fix and refactoring.

ababo:
Resolves BV-319.


proc/worker/misc.go
```diff
@@ -83,27 +83,41 @@ func (w *Worker) newUser(tx *types.Transaction) (*data.User, error) {
...
 	}
 
+	now := time.Now()
+
 	return w.queue.Add(&data.Job{
 		RelatedType: rType,
 		RelatedID:   rID,
 		Type:        jType,
-		CreatedAt:   time.Now(),
+		CreatedAt:   now,
```
sofabeat: This is set in *job.q.Add. No need to specify

proc/worker/misc.go
```diff
@@ -83,27 +83,41 @@ func (w *Worker) newUser(tx *types.Transaction) (*data.User, error) {
...
 		CreatedBy:   data.JobTask,
 		Data:        data2,
+		NotBefore:   now.Add(delay),
 	})
 }
 
+func (w *Worker) addJobWithData(
```
sofabeat: It looks like part of job package, isn't it?

ababo: @sofabeat updated

PR #161, opened 2018-06-29
Refactor tokens transfer, to get transaction immediately.

ababo:
Resolves BV-449.

### DIFF
--- a/main.go
+++ b/main.go
@@ -164,7 +164,8 @@ func main() {
 	runner := svcrun.NewServiceRunner(conf.ServiceRunner, logger, db, pr)
 	worker.SetRunner(runner)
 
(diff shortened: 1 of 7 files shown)


ababo:
> Also PreAccountAddBalanceApprove seems to be redundant worker func. And It seems to me that we can just get rid of it.

It was asked to do so.

ababo:
That's why it shouldn't be moved outside unless we plan to do a serious refactoring which will touch a lot of things.

ababo:
I suggest we will agree on some serious refactoring not to mix a fixes like that with it.

ababo:
фывыфвыфвыфвыроловрыфловрлоыфрволыф

PR #171, opened 2018-07-05
add dappctrl.config.dev.json

drew2a:
I made the following changes:
1. PSC address
1. SOMC address
1. Report.ReleaseStage


ababo:
I suggest to rename the file as `dappctrl-dev.config.json` to be more consistent with `dappctrl-test.config.json`

PR #173, opened 2018-07-05
Ethereum client request timeout.

dzeckelev:
Resolves BV-540.

### DIFF
--- a/dappctrl.config.json
+++ b/dappctrl.config.json
@@ -28,7 +28,12 @@
             "PTCAddrHex": "0x0d825eb81b996c67a55f7da350b6e73bab3cb0ec",
             "PSCAddrHex": "0x7442875d5336894cf2db5f4e295f80538b7b14c0"
         },
-        "GethURL": "https://rinkeby.infura.io/k7mXdaE6eHJ4xMnOvx8Z"
+        "GethURL": "https://rinkeby.infura.io/k7mXdaE6eHJ4xMnOvx8Z",
+        "Timeout": {
+            "ResponseHeaderTimeout":20,
+            "TLSHandshakeTimeout":5,
+            "ExpectContinueTimeout":5
+        }
     },
     "Gas": {
         "PTC": {
--- a/dappctrl.config.json
+++ b/dappctrl.config.json
@@ -30,9 +30,9 @@
         },
         "GethURL": "https://rinkeby.infura.io/k7mXdaE6eHJ4xMnOvx8Z",
         "Timeout": {
-            "ResponseHeaderTimeout":20,
-            "TLSHandshakeTimeout":5,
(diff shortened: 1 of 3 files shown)


dappctrl.config.json
```diff
@@ -28,7 +28,12 @@
             "PTCAddrHex": "0x0d825eb81b996c67a55f7da350b6e73bab3cb0ec",
             "PSCAddrHex": "0x7442875d5336894cf2db5f4e295f80538b7b14c0"
         },
-        "GethURL": "https://rinkeby.infura.io/k7mXdaE6eHJ4xMnOvx8Z"
+        "GethURL": "https://rinkeby.infura.io/k7mXdaE6eHJ4xMnOvx8Z",
+        "Timeout": {
+            "ResponseHeaderTimeout":20,
+            "TLSHandshakeTimeout":5,
+            "ExpectContinueTimeout":5
```
ababo: Minor suggestion: please add a spaces between names and values.

eth/client.go
```diff
@@ -12,20 +12,49 @@ package eth
...
 )
 
 // Block labels.
 const (
 	BlockLatest = "latest"
 )
 
+var (
+	tlsConfig = &tls.Config{
+		ClientSessionCache: tls.NewLRUClientSessionCache(1024),
```
ababo: Suggest to move the magic number to a constant.

eth/client.go
```diff
@@ -12,20 +12,49 @@ package eth
...
+	GethURL  string
+	Timeout  timeout
+}
+
+type contract struct {
+	PTCAddrHex string
+	PSCAddrHex string
+}
+
+type timeout struct {
```
ababo: Do we really need that grouping?

eth/client.go
```diff
@@ -35,6 +64,58 @@ type EthereumClient struct {
...
+	httpEthClient, err := rpc.DialHTTPWithClient(conf.GethURL, httpClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return ethclient.NewClient(httpEthClient), nil
+}
+
+func newHTTPClient(conf *Config) *http.Client {
+	if conf.Timeout.ResponseHeaderTimeout == 0 {
```
ababo: You don't need to manually assign default values. Your `NewConfig()` should return a default instance, then it'll be modified by JSON parsed, i.e. only the fields which are present in the configuration files will be modified, others will remain default.

eth/client.go
dzeckelev: It seems to me that yes. We may have other groups, in the future.

eth/client.go
dzeckelev: This is not necessary here, because the user can enter zero values for the timeouts in the configuration.
In this case, the default configuration will not help.

PR #196, opened 2018-07-16
Separate product specific logic from dappctrl.

dzeckelev:
Resolves BV-571.

Changes:
* Added additional REST endpoint to `sessrv` to get endpoint object from database.
* Added additional package `prepare`  to  `dappvpn` for prepare configuration for client.
* All Product specific code in packages has been refactored and moved to `svc\dappvpn\msg`.


svc/dappvpn/prepare/prepare_test.go
```diff
@@ -0,0 +1,135 @@
...
+	adapterConfig.Monitor = conf.VPNMonitor
+
+	if err := ClientConfig(fxt.Channel.ID, adapterConfig); err != nil {
+		t.Fatal(err)
+	}
+
+	target := filepath.Join(rootDir, fxt.Endpoint.Channel)
+
+	checkFile(t, configDestination(target))
+	checkFile(t, accessDestination(target))
```
furkhat: So you only check file that file was created. Should we check file content too?

PR #204, opened 2018-07-23
schedule update current supply jobs

furkhat:


monitor/schedule.go
```diff
@@ -354,33 +363,51 @@ func (m *Monitor) findChannelID(el *data.EthLog) string {
...
 	var offering data.Offering
-	if err := data.FindOneTo(m.db.Querier, &offering, "hash", offeringHash); err != nil {
-		m.logger.Error(err.Error())
-		return
+	err := m.db.FindOneTo(&offering, "hash", offeringHash)
+	if err == sql.ErrNoRows {
+		// Try to take related id from after offering publish job.
+		jobsLog := &data.EthLog{}
+		topicHex := topic.Hex()
+		for len(topicHex) < 64 {
```
sofabeat: Not sure about efficiency, but maybe 
* `len(topicHex) <= 64` to remove line 379 OR
* fmt.Sprintf("%064v", topicHex) OR
* use SQL query with regex pattern matching

monitor/schedule.go
```diff
@@ -354,33 +363,51 @@ func (m *Monitor) findChannelID(el *data.EthLog) string {
...
+			topicHex = "0" + topicHex
+		}
+		topicHex = "0" + topicHex
+		err = m.db.SelectOneTo(jobsLog, `WHERE topics->>2 = $1
+						AND type IN ($2, $3)`, topicHex,
+			data.JobAgentAfterOfferingMsgBCPublish,
```
sofabeat: it looks, that `clientUpdateCurrentSupplySchedulers` can be triggered only by Client, based on `if agent {..} else {...}`  (line 105), thus it cannot be `JobAgentAfterOfferingMsgBCPublish`, but only `JobClientAfterOfferingMsgBCPublish`

monitor/schedule.go
```diff
@@ -354,33 +363,51 @@ func (m *Monitor) findChannelID(el *data.EthLog) string {
...
+
+	j := &data.Job{
+		Type:        jobType,
+		RelatedID:   relID,
+		RelatedType: data.JobOffering,
+		Data:        []byte("{}"),
+		CreatedBy:   data.JobBCMonitor,
+	}
+
+	if err := m.queue.Add(j); err != nil {
```
sofabeat: Why not to use `scheduleCommon` ?

monitor/schedule.go
```diff
@@ -354,33 +363,51 @@ func (m *Monitor) findChannelID(el *data.EthLog) string {
...
+	// First try to take related id from offeing in db.
+	topic := el.Topics[3]
+	offeringHash := data.FromBytes(topic.Bytes())
 	var offering data.Offering
-	if err := data.FindOneTo(m.db.Querier, &offering, "hash", offeringHash); err != nil {
-		m.logger.Error(err.Error())
-		return
+	err := m.db.FindOneTo(&offering, "hash", offeringHash)
+	if err == sql.ErrNoRows {
+		// Try to take related id from after offering publish job.
```
sofabeat: where do you assign `relID` , if found?

monitor/schedule.go
furkhat: I prefer exact matching, option `fmt.Sprintf("%064v", topicHex)` looks good, thanks

monitor/schedule.go
furkhat: But we can change our role from client to agent, or from agent to client? Should I ignore this case?

monitor/schedule.go
furkhat: First I don't want to update job id in eth log. Second is the way `scheduleCommon` handles errors, I don't want to mark event for ignore.

I understand that this is a design breaker, because initially it was intended to have single job for single eth log. I hope we'll figure some better design soon.

monitor/schedule.go
furkhat: in `else` block, line 390

monitor/schedule.go
```diff
@@ -354,33 +363,49 @@ func (m *Monitor) findChannelID(el *data.EthLog) string {
...
+			relID = *jobsLog.JobID
+		}
+	} else {
+		relID = offering.ID
 	}
-	offering.CurrentSupply = updateCurSupply(offering.CurrentSupply)
-	if err := data.Save(m.db.Querier, &offering); err != nil {
-		m.logger.Error(err.Error())
+
+	if relID == "" {
```
ababo: Empty `reID` can mean that something wrong was happening with the DB connection (the case when `err != nil && err != sql.ErrNoRows`).

PR #208, opened 2018-07-27
Added permissions for settings.

dzeckelev:
Resolves BV-409, BV-658, BV-659.

### DIFF
--- a/uisrv/setting.go
+++ b/uisrv/setting.go
@@ -3,6 +3,8 @@ package uisrv
 import (
 	"net/http"
 
+	"gopkg.in/reform.v1"
+
 	"github.com/privatix/dappctrl/data"
 )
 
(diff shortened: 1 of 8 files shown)


uisrv/setting.go
```diff
@@ -43,6 +47,29 @@ func (s *Server) handlePutSettings(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	for _, setting := range payload {
+		var settingFromDB data.Setting
+
+		// gets setting from database
+		if err := tx.FindByPrimaryKeyTo(&settingFromDB,
+			setting.Key); err != nil {
+			if err != reform.ErrNoRows {
+				tx.Rollback()
```
ababo: I suggest to use the following approach to minimise explicit rollbacks:
```go
// Within transaction.
defer tx.Rollback()

// Some flow with bare returns on error.

tx.Commit()
// Here we return with success.
```

uisrv/setting.go
```diff
@@ -43,6 +47,29 @@ func (s *Server) handlePutSettings(w http.ResponseWriter, r *http.Request) {
...
+				s.replyUnexpectedErr(w)
+				return
+			}
+		}
+
+		// if setting exists and settings.permissions < 2
+		// then setting ignored
+		if settingFromDB.Key != "" {
+			if setting.Permissions < 2 {
```
furkhat: Use constants you defined in `schema.go`

PR #212, opened 2018-07-31
write app version

furkhat:


tool/dappinst/main.go
```diff
@@ -133,3 +135,15 @@ func createDappvpnConfig(logger *util.Logger,
 		logger.Fatal("failed to write dappvpn config")
 	}
 }
+
+func writeAppVersion(logger *util.Logger, db *reform.DB) {
```
sofabeat: Let's write to log:
1. Old version (before update)
2. New version (after update)

furkhat: how I get version before update?

furkhat: Is this script meant to be run on non-clean db?

sofabeat: This script runs on non-clean DB. Main goal of this this script to set unique authentication parameters both in DB and config files of corresponding service adapters (currently dappvpn).

PR #214, opened 2018-07-31
change logs names

drew2a:


dappctrl-dev.config.json
```diff
@@ -177,14 +177,14 @@
             "VPN server": {
                 "Name": "/root/go/bin/dappvpn",
                 "Args": [
-                    "-config=/opt/privatix/config/dappvpn.config.json"
+                    "-config=/opt/privatix/config/dappvpn-server.config.json"
```
kelewind: This config can not now be renamed, since it
1) It is dynamically generated
2) It has a part of the installer's logic

kelewind:
@drew2a 
All renaming of config files through tasks, it is necessary to plan for the next sprint
At the moment, if you want to change the name - the maximum name of the logs

PR #223, opened 2018-08-04
Refactor logging for prepare and msg pkgs.

dzeckelev:
Resolves BV-617.

### DIFF
--- a/svc/dappvpn/main.go
+++ b/svc/dappvpn/main.go
@@ -6,7 +6,6 @@ import (
 	"flag"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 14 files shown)


svc/dappvpn/main.go
```diff
@@ -19,12 +19,14 @@ import (
 	"github.com/privatix/dappctrl/svc/dappvpn/msg"
 	"github.com/privatix/dappctrl/svc/dappvpn/prepare"
 	"github.com/privatix/dappctrl/util"
+	"github.com/privatix/dappctrl/util/log"
 )
```
furkhat: I've seen source files where we name our log as `log2` and default log package as it is. Lets follow that convention in all files?

svc/dappvpn/msg/msg_test.go
```diff
@@ -13,20 +13,23 @@ import (
 	"github.com/privatix/dappctrl/sesssrv"
 	"github.com/privatix/dappctrl/svc/dappvpn/mon"
 	"github.com/privatix/dappctrl/util"
+	"github.com/privatix/dappctrl/util/log"
```
furkhat: I've seen source files where we name our log as `log2` and default log package as it is. Lets follow that convention in all files?

svc/dappvpn/main.go
dzeckelev: I deleted the default log package.

svc/dappvpn/msg/msg_test.go
dzeckelev: I deleted the default log package.

PR #243, opened 2018-08-25
Replaced GetOfferingSupply() with GetOfferingInfo().

dzeckelev:
Resolves BV-715.

### DIFF
--- a/proc/worker/adapter.go
+++ b/proc/worker/adapter.go
@@ -45,8 +45,10 @@ type EthBackend interface {
 
 	PSCReturnBalanceERC20(*bind.TransactOpts, *big.Int) (*types.Transaction, error)
 
-	PSCOfferingSupply(opts *bind.CallOpts,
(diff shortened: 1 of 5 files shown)


proc/worker/adapter.go
```diff
@@ -215,18 +217,22 @@ func (b *ethBackendInstance) PSCAddBalanceERC20(opts *bind.TransactOpts,
 	return tx, nil
 }
 
-func (b *ethBackendInstance) PSCOfferingSupply(
-	opts *bind.CallOpts, hash [common.HashLength]byte) (uint16, error) {
+func (b *ethBackendInstance) GetOfferingInfo(opts *bind.CallOpts,
```
furkhat: it seems like there was a convention to prefix methods with related contract.
`PSCGetOfferingInfo`

dzeckelev: updated

PR #253, opened 2018-08-30
database migration: create database schema and fill data

ubozov:


ubozov:
updated commit contains:
a. dappctrl self check when started to run new migration scripts 
b. sql files embedding via statik

ubozov:
@furkhat :  in the proposed version, when the migration scripts execute the dappctrl itself, the option to run fill_data.sql separately will not work.
installer, in that case it will not be able to run it separately, because at the time of installation the dappctrl has not been started yet and the launch of fill_data.sql will throw an exception (schema has not created yet).

furkhat:
I'm worried about having initial database configuration spread over multiple files/migrations (which will eventually happen if we put it all into the migrations). It is much easier and (maybe?) better to have it as before, single (`prod_data.sql`) sql file.

data/command.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+
+	"github.com/privatix/dappctrl/data/migration"
+	"github.com/privatix/dappctrl/statik"
+)
+
+// ExecuteCommand does execute commands to manage database
+// db-update - command for execute migration scripts
+// db-init - command for initialize database by default values.
+func ExecuteCommand(args []string) error {
```
furkhat: Let's change command names, right now I hardly can tell, just by seeing names, which one runs migrations.
I propose:
```
// ExecuteCommand executes commands to manage database
// db-migrate - command for execute migration scripts
// db-init-data - command for initialize database by default values.
...
```

data/command.go
```diff
@@ -0,0 +1,78 @@
...
+	if len(args) == 0 {
+		return nil
+	}
+
+	switch args[0] {
+	case "db-update":
+		connStr := getConnectionString(args)
```
furkhat: Lets get connection string as we used to. Though config file. We could just pass `DBConfig` instead of args to `ExecuteCommand`

data/command.go
```diff
@@ -0,0 +1,78 @@
...
+		if err := initData(connStr); err != nil {
+			panic(fmt.Sprintf("failed to init database %s", err))
+		}
+		os.Exit(1)
+	}
+	return nil
+}
+
+func getConnectionString(args []string) string {
+	connStr := flag.String("conn", "", "Database connection string")
```
furkhat: If you use `DBConfig`, you can just use `ConnStr()` method.

data/command.go
ubozov: ok

data/command.go
ubozov: for installer it is more comfortable to use a string instead of a config file

data/command.go
ubozov: for installer it is more comfortable to use a string instead of a config file

data/command.go
furkhat: ok. Could you leave an example in Readme file?

data/schema.sql
```diff
@@ -355,4 +355,16 @@ CREATE TABLE log_events (
...
+CREATE TABLE goose_db_version (
+            	id serial NOT NULL,
+                version_id bigint NOT NULL,
+                is_applied boolean NOT NULL,
+                tstamp timestamp NULL default now(),
+                PRIMARY KEY(id)
+            );
+
+INSERT INTO goose_db_version (version_id, is_applied) VALUES (0, true);
+INSERT INTO goose_db_version (version_id, is_applied) VALUES (1, true);
```
furkhat: Do we need this file now anyway? Doesn't migration files suffice?

data/command.go
```diff
@@ -0,0 +1,78 @@
...
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/privatix/dappctrl/data/migration"
+	"github.com/privatix/dappctrl/statik"
+)
+
+// ExecuteCommand does execute commands to manage database
```
furkhat: `ExecuteComand exectes ...`

data/schema.sql
ubozov: this file was removed and readme file was updated

data/command.go
ubozov: fixed

PR #259, opened 2018-09-03
Move templates endpoints to a JSON RPC server.

dzeckelev:
Resolves BV-763.

### DIFF
--- a/doc/ui/rpc.md
+++ b/doc/ui/rpc.md
@@ -96,6 +96,51 @@ This document describes a UI JSON RPC API located in "ui" namespace.
 *Result*: None.
 
 
+### Objects
+
+#### Get object
+
+*Method*:	`getObject`
+
+*Description*: Get an object of a specified type..
+
+*Parameters*:
+1. Password (string)
+2. Object type (string, can be `account`, `user`, `template`, `product`,
+ `offering`, `channel`, `session`, `contract`, `endpoint`, `job`, `ethTx` or `ethLog`)
+3. Object id (string)
+
+*Result (object)*:
+- `object` (object) - object of a given type.
+
+### Templates
+
+#### Create Template
+
+*Method*:	`createTemplate`
+
+*Description*: Create new template.
+
+*Parameters*:
+1. Password (string)
+2. Template (data.Template object)
+
+*Result (object)*:
+- `template` (string) - id of template to be created.
+
+#### Get Templates
+
+*Method*:	`getTemplates`
+
+*Description*: Get templates.
+
+*Parameters*:
+1. Password (string)
+2. Template type (string, can be `offer` or `access`)
+
+*Result (object)*:
+- `templates` (array of `data.Template` objects) - returned templates.
+
 ## Subscriptions to asynchronous notifications
 
 #### Object change
--- a/ui/db.go
+++ b/ui/db.go
@@ -1,8 +1,9 @@
 package ui
 
 import (
+	"gopkg.in/reform.v1"
+
 	"github.com/privatix/dappctrl/util/log"
-	reform "gopkg.in/reform.v1"
 )
 
 func (h *Handler) findByPrimaryKey(logger log.Logger,
@@ -27,6 +28,29 @@ func (h *Handler) findByColumn(logger log.Logger, notFoundError error,
 		return ErrInternal
 	}
 	return nil
+}
+
+func (h *Handler) selectOneTo(logger log.Logger,
+	notFoundError error, record reform.Record,
+	tail string, args ...interface{}) error {
+	if err := h.db.SelectOneTo(record, tail, args...); err != nil {
+		logger.Error(err.Error())
+		if err == reform.ErrNoRows {
+			return notFoundError
+		}
+		return ErrInternal
+	}
+	return nil
+}
+
+func (h *Handler) selectAllFrom(logger log.Logger, view reform.View,
+	tail string, args ...interface{}) ([]reform.Struct, error) {
+	rows, err := h.db.SelectAllFrom(view, tail, args...)
+	if err != nil {
+		logger.Error(err.Error())
+		return nil, ErrInternal
+	}
+	return rows, err
 }
 
 func beginTX(logger log.Logger, db *reform.DB) (*reform.TX, error) {
--- a/ui/object.go
+++ b/ui/object.go
@@ -1,0 +1,88 @@
+package ui
+
+import (
+	"encoding/json"
+
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+)
+
+// Objects types.
+const (
+	TypeAccount  = "account"
+	TypeUser     = "user"
+	TypeTemplate = "template"
+	TypeProduct  = "product"
+	TypeOffering = "offering"
+	TypeChannel  = "channel"
+	TypeSession  = "session"
+	TypeContract = "contract"
+	TypeEndpoint = "endpoint"
+	TypeJob      = "job"
+	TypeEthTx    = "ethTx"
+	TypeEthLog   = "ethLog"
+)
+
+var objectTypes = map[string]reform.Table{
+	TypeAccount:  data.AccountTable,
+	TypeUser:     data.UserTable,
+	TypeTemplate: data.TemplateTable,
+	TypeProduct:  data.ProductTable,
+	TypeOffering: data.OfferingTable,
+	TypeChannel:  data.ChannelTable,
+	TypeSession:  data.SessionTable,
+	TypeContract: data.ContractTable,
+	TypeEndpoint: data.EndpointTable,
+	TypeJob:      data.JobTable,
+	TypeEthTx:    data.EthTxTable,
+	TypeEthLog:   data.EthLogTable,
+}
+
+// GetObjectResult is a GetObject result.
+type GetObjectResult struct {
+	Object json.RawMessage `json:"object"`
+}
+
+// GetObject finds object in a database by id,
+// then returns an object on raw JSON format.
+func (h *Handler) GetObject(
+	password, objectType, id string) (*GetObjectResult, error) {
+	logger := h.logger.Add("method", "GetObject",
+		"type", objectType, "id", id)
+
+	err := h.checkPassword(logger, password)
+	if err != nil {
+		return nil, err
+	}
+
+	table, ok := objectTypes[objectType]
+	if !ok {
+		logger.Warn(ErrBadObjectType.Error())
+		return nil, ErrBadObjectType
+	}
+
+	obj, err := h.db.FindByPrimaryKeyFrom(table, id)
+	if err != nil {
+		logger.Error(err.Error())
+		return nil, ErrObjectNotFound
+	}
+
+	raw, err := json.Marshal(obj)
+	if err != nil {
+		logger.Error(err.Error())
+		return nil, ErrInternal
+	}
+
+	return &GetObjectResult{raw}, nil
+}
+
+func (h *Handler) insertObject(object reform.Struct) error {
+	logger := h.logger.Add("method", "insertObject", "object", object)
+
+	if err := h.db.Insert(object); err != nil {
+		logger.Error(err.Error())
+		return ErrInternal
+	}
+	return nil
+}
--- a/ui/template.go
+++ b/ui/template.go
@@ -1,0 +1,97 @@
+package ui
+
+import (
+	"encoding/json"
+
+	"github.com/ethereum/go-ethereum/crypto"
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/util"
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// CreateTemplateResult is a CreateTemplate result.
+type CreateTemplateResult struct {
+	Template string `json:"template"`
+}
+
+// GetTemplatesResult is a GetTemplates result.
+type GetTemplatesResult struct {
+	Templates []data.Template `json:"templates"`
+}
+
+// GetTemplates returns templates.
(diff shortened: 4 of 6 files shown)


ui/template.go
```diff
@@ -0,0 +1,102 @@
...
+	if !ok {
+		logger.Error("invalid result type")
+		return nil, ErrInternal
+	}
+
+	var templates []*data.Template
+
+	for _, v := range rows {
+		template := v.(*data.Template)
+		templates = append(templates, template)
```
furkhat: Why do you need to convert `result`? What if you just return `[]reform.Struct`? Isn't it should work the same?

ui/generic.go
```diff
@@ -0,0 +1,173 @@
...
+		cond := fmt.Sprintf("%s %s %s", param.Field, op, ph)
+		conds = append(conds, cond)
+	}
+
+	return conds, args
+}
+
+// handleGetResources select and returns records.
+func (h *Handler) handleGetResources(params map[string]string,
+	conf *getConf) (interface{}, error) {
```
furkhat: Let's return `[]reform.Struct` instead.

dzeckelev: here returns not  `[]reform.Struct` https://github.com/Privatix/dappctrl/blob/d5a02eee18cfec69ecc011f60ebe9174eab88505/ui/generic.go#L169

ui/template.go
dzeckelev: this code is blocking me https://github.com/Privatix/dappctrl/blob/d5a02eee18cfec69ecc011f60ebe9174eab88505/ui/generic.go#L169

dzeckelev: Right. `[]reform.Struct` is the same as `[]*data.Template`.
Updated.

ui/db.go
```diff
@@ -0,0 +1,17 @@
...
+	"fmt"
+
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/util/log"
+)
+
+func (h *Handler) insert(logger log.Logger, rec reform.Struct) error {
+	if err := h.db.Insert(rec); err != nil {
+		logger.Error(fmt.Sprintf("failed to insert: %v", err))
```
ababo: Please log record type, i.e. "failed to insert %T: %s".

ui/template.go
```diff
@@ -0,0 +1,95 @@
...
+// CreateTemplateResult is a CreateTemplate result.
+type CreateTemplateResult struct {
+	Template string `json:"template"`
+}
+
+// TemplatesResult is a Templates result.
+type TemplatesResult []reform.Struct
+
+// Templates returns templates.
+func (h *Handler) Templates(
```
ababo: Please name synchronous methods using verbs, e.g. `DoSomething()`.

ui/generic.go
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		cond := fmt.Sprintf("%s %s %s", param.Field, op, ph)
+		conds = append(conds, cond)
+	}
+
+	return conds, args
+}
+
+// handleGetResources select and returns records.
```
ababo: selects

ababo: Please write a detailed comment explaining what does this function do. Currently I doubt we need such a complex logic here.

ui/template_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"testing"
+
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/util"
+)
+
+func testGetTemplates(t *testing.T, exp int,
+	id, templateType string, err error) TemplatesResult {
```
ababo: Better to rename into `expected`.

ui/template_test.go
```diff
@@ -0,0 +1,79 @@
...
+	defer fxt.close()
+
+	_, err := handler.CreateTemplate(
+		"wrong-password", *fxt.TemplateOffer)
+	util.TestExpectResult(t, "CreateTemplate", ErrAccessDenied, err)
+
+	for _, template := range []data.Template{
+		{
+			Kind: data.TemplateOffer,
+			Raw:  []byte("{\"fake\" : \"" + util.NewUUID() + "\"}"),
```
ababo: Better to use raw literals.

ui/template_test.go
```diff
@@ -0,0 +1,79 @@
...
+	} {
+		res, err := handler.CreateTemplate(data.TestPassword, template)
+		util.TestExpectResult(t, "CreateTemplate", nil, err)
+
+		tpl := &data.Template{}
+		err = db.FindByPrimaryKeyTo(tpl, res.Template)
+		if err != nil {
+			t.Fatal(err)
+		}
+		data.DeleteFromTestDB(t, db, tpl)
```
ababo: This should be deferred not to spoil the database when the test fails.

dzeckelev: If the test fails, a template will not be in database. This is checked in the previous step.

ui/template.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// CreateTemplateResult is a CreateTemplate result.
+type CreateTemplateResult struct {
+	Template string `json:"template"`
+}
+
+// TemplatesResult is a GetTemplates result.
+type TemplatesResult []reform.Struct
```
ababo: Please tend to make RPC results as structures for flexibility (we'll be able to add new fields transparently without touching existing code).

ui/template.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	return result.Items, nil
+}
+
+func invalidTemplateKind(v string) bool {
+	return v != data.TemplateOffer && v != data.TemplateAccess
+}
+
+func isValidTemplate(logger log.Logger, template data.Template) error {
```
ababo: "Is"-functions should always return `bool` or `(bool, error)`. Please consider renaming it into `checkTemplate`.

ui/template.go
```diff
@@ -0,0 +1,87 @@
...
+	v := make(map[string]interface{})
+	if invalidTemplateKind(template.Kind) ||
+		json.Unmarshal(template.Raw, &v) != nil {
+		logger.Error("invalid template")
```
ababo: I would differentiate these two cases: there're completely different by nature. Moreover it's worthy to log errors returned.

ui/template_test.go
```diff
@@ -0,0 +1,82 @@
...
+import (
+	"testing"
+
+	"fmt"
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/util"
+)
+
+func expectedTemplates(t *testing.T, expected int,
+	id, templateType string, err error) TemplatesResult {
```
ababo: consider replacing `err`->`expected`.

ui/template_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+import (
+	"testing"
+
+	"fmt"
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/util"
+)
+
+func expectedTemplates(t *testing.T, expected int,
```
ababo: You never check template content. Would suggest to do that.

ui/template.go
```diff
@@ -0,0 +1,118 @@
...
+		"method", "GetTemplates", "id", id, "type", templateType)
+
+	err := h.checkPassword(logger, password)
+	if err != nil {
+		return nil, err
+	}
+
+	result := &TemplatesResult{}
+
+	if id != "" {
```
ababo: Where do we need a scenario when all the templates are retrieved at once?

ui/template.go
```diff
@@ -0,0 +1,118 @@
+package ui
+
+import (
```
ababo: Please consider having a generic call when you pass object type and id to get an instance. That way we'll avoid lot of code duplications. See my subscriptions as example.

ui/template.go
dzeckelev: https://github.com/Privatix/dappctrl/blob/6b2e3bc864d1200cd9e50ace0715b738ec969391/ui/template.go#L64

ui/template.go
dzeckelev: @ababo updated.

ui/object.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+)
+
+// Objects types.
+const (
+	AccountType  = "account"
```
ababo: TypeAccount, TypeUser, etc.

ui/object.go
```diff
@@ -0,0 +1,75 @@
...
+	EndpointType: data.EndpointTable,
+	JobType:      data.JobTable,
+	EthTxType:    data.EthTxTable,
+	EthLogType:   data.EthLogTable,
+}
+
+// GetObject returns an object of the specified type.
+// Type must implement the "reform.Struct" interface.
+func (h *Handler) GetObject(
+	password, id, objectType string) (result reform.Struct, err error) {
```
ababo: Please return `json.RawMessage` instead (otherwise RPC client won't know how to deserialize the result).

ui/template.go
```diff
@@ -0,0 +1,100 @@
...
+	Template string `json:"template"`
+}
+
+// TemplatesResult is a GetTemplates result.
+type TemplatesResult struct {
+	Templates []reform.Struct `json:"templates"`
+}
+
+// GetTemplates returns templates.
+func (h *Handler) GetTemplates(
```
ababo: Why didn't you remove id here?

ui/object.go
dzeckelev: Client correctly perceives the result. I checked with Postman tool.

ababo: Try with RPC client.

ui/template.go
dzeckelev: @ababo updated.

ui/template.go
```diff
@@ -0,0 +1,104 @@
...
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// CreateTemplateResult is a CreateTemplate result.
+type CreateTemplateResult struct {
+	Template string `json:"template"`
+}
+
+// TemplatesResult is a GetTemplates result.
+type TemplatesResult struct {
```
ababo: GetTemplatesResult

ui/template.go
```diff
@@ -0,0 +1,104 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	result := &GetTemplatesResult{}
+
+	for _, v := range templates {
+		template, ok := v.(*data.Template)
+		if !ok {
```
ababo: no need for that, just `append(result.Templates, *v.(*data.Template))`

ui/template.go
```diff
@@ -0,0 +1,104 @@
...
+			continue
+		}
+		result.Templates = append(result.Templates, *template)
+	}
+
+	return result, nil
+}
+
+func invalidTemplateKind(v string) bool {
```
ababo: Do we need that function? Why not to inline it?

doc/ui/rpc.md
```diff
@@ -33,6 +33,52 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+*Method*:	`getTemplates`
+
+*Description*: Get templates.
+
+*Parameters*:
+1. Password (string)
+2. Template type (string, can be `offer` or `access`)
+
+*Result (object)*:
+- `templates` (array of `data.Template` objects) - templates objects.
```
ababo: ) - returned templates

ui/template.go
```diff
@@ -0,0 +1,100 @@
...
+func checkTemplate(logger log.Logger, kind string, raw []byte) error {
+	v := make(map[string]interface{})
+	if kind != data.TemplateOffer &&
+		kind != data.TemplateAccess {
+		logger.Error("invalid template type")
+		return ErrInvalidTemplateType
+	}
+	if err := json.Unmarshal(raw, &v); err != nil {
+		logger.Error(err.Error())
+		return ErrInvalidTemplateRawData
```
ababo: ErrMalformedTemplate

ui/template.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	return nil
+}
+
+// CreateTemplate creates template.
+func (h *Handler) CreateTemplate(password, kind string,
+	raw []byte) (*CreateTemplateResult, error) {
```
ababo: Just pass a pointer to `data.Template` instead.

dzeckelev: @ababo updated.

ui/misc.go
```diff
@@ -30,6 +30,29 @@ func (h *Handler) findByPrimaryKey(logger log.Logger,
...
+		logger.Error(err.Error())
+		if err == reform.ErrNoRows {
+			return notFoundError
+		}
+		return ErrInternal
+	}
+	return nil
+}
+
+func (h *Handler) selectAllFrom(logger log.Logger, view reform.View,
```
ababo: Please move the method to db.go (if it doesn't exist there). Probably you just need to rebase the changes on current `develop`.

ui/template.go
```diff
@@ -0,0 +1,97 @@
...
+		templates, err = h.selectAllFrom(
+			logger, data.TemplateTable,
+			"WHERE kind = $1", tplType)
+	} else {
+		templates, err = h.selectAllFrom(
+			logger, data.TemplateTable, "")
+	}
+
+	if err != nil {
+		return nil, err
```
ababo: You cannot return errors from reform, only errors of our type. So please log the error and return `ErrInternal`. Please recheck all your returns to make sure we return only `errors.Error` instances.

dzeckelev: All errors processed in `selectAllFrom`, `selectAllFrom`. These methods return `errors.Error`. We return our errors.
Why again catch errors?

ababo: Oh, you're right...

PR #260, opened 2018-09-04
Move settings endpoints to a JSON RPC server.

dzeckelev:
Resolves BV-769.

### DIFF
--- a/doc/ui/rpc.md
+++ b/doc/ui/rpc.md
@@ -16,9 +16,34 @@ This document describes a UI JSON RPC API located in "ui" namespace.
 3. Offering id (string)
 4. Gas price (number)
 
-*Result*:
+*Result (object)*:
 - `channel` (string) - id of channel to be created.
 
+### Settings
+
+#### Get Settings
+
+*Method*:	`getSettings`
+
+*Description*: Get settings.
+
+*Parameters*:
+1. Password (string)
+
+*Result (object)*:
+- `settings` (object) - object with keys as setting names and values as setting values.
+
+#### Update Settings
+
+*Method*:	`updateSettings`
+
+*Description*: Update existing settings.
+
+*Parameters*:
+1. Password (string)
+2. Object with keys as setting names and values as setting values (object)
+
+*Result*: None.
 
 #### Create Product
 
@@ -30,7 +55,7 @@ This document describes a UI JSON RPC API located in "ui" namespace.
 1. Password (string)
 2. Product (`data.Product` object)
 
-*Result*:
+*Result (object)*:
 - `product` (string) - id of created product.
 
 
@@ -42,7 +67,7 @@ This document describes a UI JSON RPC API located in "ui" namespace.
 
 *Parameters*: None.
 
-*Result*:
+*Result (object)*:
 1. Products (array of data.Product objects)
 
 
@@ -123,7 +148,7 @@ This document describes a UI JSON RPC API located in "ui" namespace.
 
 *Parameters*:
 1. Password (string)
-2. Template (data.Template object)
+2. Template (`data.Template` object)
 
 *Result (object)*:
 - `template` (string) - id of template to be created.
--- a/ui/db.go
+++ b/ui/db.go
@@ -30,19 +30,6 @@ func (h *Handler) findByColumn(logger log.Logger, notFoundError error,
 	return nil
 }
 
-func (h *Handler) selectOneTo(logger log.Logger,
-	notFoundError error, record reform.Record,
-	tail string, args ...interface{}) error {
-	if err := h.db.SelectOneTo(record, tail, args...); err != nil {
-		logger.Error(err.Error())
-		if err == reform.ErrNoRows {
-			return notFoundError
-		}
-		return ErrInternal
-	}
-	return nil
-}
-
 func (h *Handler) selectAllFrom(logger log.Logger, view reform.View,
 	tail string, args ...interface{}) ([]reform.Struct, error) {
 	rows, err := h.db.SelectAllFrom(view, tail, args...)
--- a/ui/misc.go
+++ b/ui/misc.go
@@ -6,6 +6,7 @@ import (
 	"gopkg.in/reform.v1"
 
 	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/util/errors"
 	"github.com/privatix/dappctrl/util/log"
 )
 
@@ -81,4 +82,12 @@ func (h *Handler) defaultGasPrice(logger log.Logger) (uint64, error) {
 	}
 
 	return val, nil
+}
+
+func (h *Handler) catchError(logger log.Logger, err error) error {
+	if e, ok := err.(errors.Error); ok {
+		return e
+	}
+	logger.Error(err.Error())
+	return ErrInternal
 }
--- a/ui/settings.go
+++ b/ui/settings.go
@@ -1,0 +1,87 @@
+package ui
+
+import (
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+)
+
+const settingsCondition = "WHERE permissions > 0"
+
+// GetSettingsResult is a GetSettings result.
+type GetSettingsResult struct {
+	Settings map[string]string `json:"settings"`
+}
+
+// GetSettings returns settings.
+func (h *Handler) GetSettings(
+	password string) (*GetSettingsResult, error) {
+	logger := h.logger.Add("method", "GetSettings")
+
+	err := h.checkPassword(logger, password)
+	if err != nil {
+		return nil, err
+	}
+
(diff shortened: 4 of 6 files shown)


ui/settings.go
```diff
@@ -0,0 +1,108 @@
...
+		err = tx.FindByPrimaryKeyTo(&settingFromDB,
+			setting.Key)
+		if err != nil {
+			if err != reform.ErrNoRows {
+				logger.Error(err.Error())
+				return nil, ErrInternal
+			}
+		}
+
+		// if setting exists and settings.permissions != 2
```
furkhat: Why break abstraction. Replace `2` with `data.ReadWrite`. Little thing but anyway.

ui/db.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// Transactional funcs.
+
+func (h *Handler) begin(logger log.Logger) (*reform.TX, error) {
+	tx, err := h.db.Begin()
+	if err != nil {
+		logger.Error(fmt.Sprintf(
```
ababo: Suggest to replace with `"failed to begin transaction: "+err.Error()` or just `err.Error()`.

ui/misc.go
```diff
@@ -28,6 +28,29 @@ func (h *Handler) findByPrimaryKey(logger log.Logger,
...
+			return notFoundError
+		}
+		return ErrInternal
+	}
+	return nil
+}
+
+func (h *Handler) selectAllFrom(logger log.Logger, view reform.View,
+	tail string, args ...interface{}) ([]reform.Struct, error) {
+	rows, err := h.db.SelectAllFrom(view, tail, args...)
```
ababo: Bad name (`rows`) since it can be confused with `sql.Rows`, please use `recs` instead.

ui/settings.go
```diff
@@ -0,0 +1,102 @@
...
+	"github.com/privatix/dappctrl/data"
+)
+
+const (
+	settingsCondition = "WHERE permissions > 0"
+	settingsUpdated   = "updated."
+)
+
+// GetSettingsResult is a GetSettings result.
+type GetSettingsResult []reform.Struct
```
ababo: Please use structs for flexibility. Already told you that.

ui/settings.go
```diff
@@ -0,0 +1,102 @@
...
+	logger := h.logger.Add("method", "GetSettings", "key", key)
+
+	err := h.checkPassword(logger, password)
+	if err != nil {
+		return nil, err
+	}
+
+	if key != "" {
+		var setting data.Setting
+		err = h.selectOneTo(logger, ErrSettingNotFound, &setting,
```
ababo: Does it make sense to use `data.ReadSetting`?

ui/settings.go
ababo: Please consider using `map[string]string` instead.

ui/settings.go
```diff
@@ -0,0 +1,102 @@
...
+// GetSettings returns settings.
+func (h *Handler) GetSettings(password, key string) (GetSettingsResult, error) {
+	logger := h.logger.Add("method", "GetSettings", "key", key)
+
+	err := h.checkPassword(logger, password)
+	if err != nil {
+		return nil, err
+	}
+
+	if key != "" {
```
ababo: Please consider returning all the settings at once without filtering. I doubt the request-per-each-setting scenario is reasonable.

ui/settings.go
```diff
@@ -0,0 +1,102 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return settings, err
+}
+
+// UpdateSettings updates settings.
+func (h *Handler) UpdateSettings(password string,
+	items []data.Setting) (*UpdateSettingsResult, error) {
```
ababo: I guess we should pass `map[string]string` instead. Indeed there's no need to update setting descriptions, etc. I also think we need to change the `data.Setting` structure (to drop non-localisable description, etc.) but this is out of scope of this task.

ui/db.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// Transactional funcs.
+
+func (h *Handler) begin(logger log.Logger) (*reform.TX, error) {
```
ababo: Consider using `InTransaction` instead of `Begin`/`Rollback`/`Commit`.

ui/settings.go
dzeckelev: This request can not be performed without filtering.
Here we get all the results:
https://github.com/Privatix/dappctrl/blob/2fb2a676dcd90514a6f41056593d2ee7883b0617/ui/settings.go#L41

ui/settings.go
dzeckelev: @ababo updated.

doc/ui/rpc.md
```diff
@@ -19,6 +19,33 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+2. Setting key (string)
+
+*Result (object)*:
+- `settings` (array of `data.Setting` objects) - settings objects.
+
+#### Update Settings
+
+*Method*:	`updateSettings`
+
+*Description*: Update exist settings.
```
ababo: existing

doc/ui/rpc.md
```diff
@@ -19,6 +19,33 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+
+*Parameters*:
+1. Password (string)
+2. Items maps setting keys to setting values (object)
+
+*Result (object)*:
+- `message` (string) - message about a result of the operation.
```
ababo: Did you mean error? It's always returned.

ui/settings.go
```diff
@@ -0,0 +1,110 @@
+package ui
+
+import (
+	"gopkg.in/reform.v1"
+
+	"github.com/privatix/dappctrl/data"
+)
+
+const (
```
ababo: I would suggest to remove the whole block.

ui/settings.go
```diff
@@ -0,0 +1,110 @@
...
+	SettingsUpdated = "updated."
+)
+
+// GetSettingsResult is a GetSettings result.
+type GetSettingsResult struct {
+	Settings []data.Setting `json:"settings"`
+}
+
+// UpdateSettingsResult is an UpdateSettings result.
+type UpdateSettingsResult struct {
```
ababo: Please drop this type if I understood correctly.

ui/settings.go
```diff
@@ -0,0 +1,110 @@
...
+			if err != nil {
+				logger.Error(err.Error())
+				return ErrInternal
+			}
+		}
+		return nil
+	})
+
+	if err != nil {
+		return nil, err
```
ababo: You cannot return this error directly (only our errors can be returned). I suggest to use the transaction helpers written by @furkhat.

doc/ui/rpc.md
```diff
@@ -19,6 +19,32 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+
+#### Get Settings
+
+*Method*:	`getSettings`
+
+*Description*: Get settings.
+
+*Parameters*:
+1. Password (string)
+2. Setting key (string)
```
ababo: We should be able to retrieve all the settings at once as `map[string]string`. Reading a single setting doesn't make much sense. Moreover single object can be retrieved by out generic method.

doc/ui/rpc.md
```diff
@@ -19,6 +19,32 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+
+#### Update Settings
+
+*Method*:	`updateSettings`
+
+*Description*: Update existing settings.
+
+*Parameters*:
+1. Password (string)
+2. Items maps setting keys to setting values (object)
```
ababo: Object with keys as setting names and values as setting values (object).

doc/ui/rpc.md
```diff
@@ -96,6 +122,51 @@ This document describes a UI JSON RPC API located in "ui" namespace.
...
+
+#### Create Template
+
+*Method*:	`createTemplate`
+
+*Description*: Create new template.
+
+*Parameters*:
+1. Password (string)
+2. Template (data.Template object)
```
ababo:
```
2. Template (`data.Template` object)
```

doc/ui/rpc.md
ababo: @sofabeat I think we need to reform our settings a bit. I don't see much sense having both name and key fields, why not to have just a name as PK? Moreover we don't need a description since we support localisation, so this part is going to UI's responsibility (if needed). For now I would propose just to throw non-needed fields out of `data.Setting`. Regarding RPC API I would support only two calls: `getSettings` and `updateSettings`. The first one just returns a map name->value (JSON object), the second takes such an object and apply the values (honouring permissions). What do you think?

PR #261, opened 2018-09-04
move svcrun to new logger

furkhat:


client/svcrun/runner.go
```diff
@@ -114,7 +117,8 @@ func (r *serviceRunner) Start(channel string) error {
 		return err
 	}
 
-	r.logger.Warn("service adapter for channel %s has started", channel)
+	logger.Warn(fmt.Sprintf("service adapter for channel %s has started",
```
ababo: Please use context for supplying the channel.

client/svcrun/runner.go
```diff
@@ -159,8 +165,8 @@ func (r *serviceRunner) wait(
 		stderr.Close()
 	}()
 
-	r.logger.Warn("service adapter for channel %s has exited: %v",
-		channel, cmd.Wait())
+	logger.Warn(fmt.Sprintf("service adapter for channel %s has exited: %v",
```
ababo: Same is here.

client/svcrun/runner.go
```diff
@@ -169,7 +175,7 @@ func (r *serviceRunner) wait(
 
 	_, err := r.pr.SuspendChannel(channel, data.JobServiceAdapter, false)
 	if err != nil {
-		r.logger.Warn("failed to suspend channel: %s", err)
+		logger.Warn("failed to suspend channel: " + err.Error())
```
ababo: I suggest to add "error" to the context instead of formatting.

ababo: No, this is a bad idea, let's ignore it.

furkhat:
@ababo no. We still have `svc/dappvpn/mon` and `dappinst`.

PR #267, opened 2018-09-06
client channel top up invoker

furkhat:


ui/export_test.go
```diff
@@ -0,0 +1,7 @@
+package ui
+
+import "github.com/privatix/dappctrl/job"
+
+func (h *Handler) SetMockQueue(q job.Queue) {
```
ababo: Do we really need this function? If you insist please rename with `test.go` with a corresponding build tag.

ui/errors.go
```diff
@@ -10,16 +10,20 @@ const (
...
-	ErrAccessDenied:     "access denied",
-	ErrInternal:         "internal server error",
-	ErrAccountNotFound:  "account not found",
-	ErrOfferingNotFound: "offering not found",
-	ErrBadObjectType:    "bad object type",
+	ErrAccessDenied:            "access denied",
+	ErrInternal:                "internal server error",
+	ErrAccountNotFound:         "account not found",
+	ErrChannelNotFound:         "channel not found",
+	ErrDefailtGasPriceNotFound: "default gas price settings not found",
```
ababo: Typo

ui/handler_test.go
```diff
@@ -1,4 +1,4 @@
-package ui
+package ui_test
```
ababo: Why different package?

furkhat: Isolation of tests. I think it is generally better.

ui/export_test.go
furkhat: It is a conventional way to export something for tests. Filename is also a convention.

ababo: Didn't know about that option, agree, that's better.

ababo: Can you move other ui/ tests to the test package as well?

furkhat: I already did that. But only this package.

furkhat:
@simon updated.

PR #268, opened 2018-09-07
fix challenge_period

drew2a:
@ubozov 
Is this enough for the migration?
Do I need to fix this value elsewhere?


ubozov:
@drew2a , for prod_data.sql it is enough

PR #270, opened 2018-09-08
Agent notifies about channel termination.

dzeckelev:
Resolves BV-641.

### DIFF
--- a/pay/pay_test.go
+++ b/pay/pay_test.go
@@ -180,26 +180,33 @@ func TestServiceTerminate(t *testing.T) {
 	defer data.CleanTestDB(t, testDB)
 	fxt := newFixture(t)
 
-	fxt.Channel.ServiceStatus = data.ServiceTerminated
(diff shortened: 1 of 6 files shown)


pay/pay_test.go
```diff
@@ -171,10 +176,37 @@ func TestInvalidPayments(t *testing.T) {
...
+	payload := newTestPayload(t,
+		100, fxt.Channel, fxt.Offering, fxt.UserAcc)
+	err := postPayload(testDB, fxt.Channel.ID, payload, false, 0, pr)
+
+	// We can not simulate a conditions in this test
+	// when there are 2 channels for Client and Agent
+	// with different statuses of the service.
+	// Success is when the service
+	// can not be switched to the status of terminated.
+	if err == nil && err != proc.ErrBadServiceStatus {
```
furkhat: Very implicit test. Better to mock payment receiver server and returns data.ServiceTerminated, then check that channel status changed to terminated.

PR #279, opened 2018-09-12
file log location

furkhat:


util/log/file_logger.go
```diff
@@ -23,11 +25,31 @@ func NewFileConfig() *FileConfig {
...
+// LocationConfig is logs destination file configuration.
+type LocationConfig struct {
+	FilePrefix string
+	Path       string
+}
+
+// NewLocationConfig creates a new FileLogger configuration.
+func NewLocationConfig() *LocationConfig {
+	return &LocationConfig{
+		Path: "/var/log/",
```
ababo: This is not portable, especially for Windows.

util/log/file_logger.go
```diff
@@ -23,11 +25,31 @@ func NewFileConfig() *FileConfig {
 	}
 }
 
+// LocationConfig is logs destination file configuration.
+type LocationConfig struct {
```
ababo: Why do we need to have this functionality here in this package? We need to reuse it from different binaries? Why not to simply redirect `stderr` to file?

PR #280, opened 2018-09-12
Assign bandwidth limit on client VPN connections

ababo:
This is an initial implementation of traffic control for the client VPN sessions. Currently only download rate is constrained and on Linux only.

This commit includes the following changes:
1. The start session call now returns an offering.
2. Added a new `tc` package for traffic control.
3. Dappvpn calls SetRateLimit/UnsetRateLimit.

Resolves BV-786.

### DIFF
--- a/sesssrv/start.go
+++ b/sesssrv/start.go
@@ -18,6 +18,11 @@ type StartArgs struct {
 	ClientPort uint16 `json:"clientPort"`
 }
 
+// StartResult is a result of session starting.
+type StartResult struct {
+	Offering *data.Offering
+}
+
 func (s *Server) handleStart(
 	w http.ResponseWriter, r *http.Request, ctx *srv.Context) {
 	logger := s.logger.Add("method", "handleStart", "sender", r.RemoteAddr)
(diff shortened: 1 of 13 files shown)


sesssrv/start.go
```diff
@@ -36,6 +41,13 @@ func (s *Server) handleStart(
 		return
 	}
 
+	var offer data.Offering
+	if err := s.db.FindByPrimaryKeyTo(&offer, ch.Offering); err != nil {
+		logger.Error(err.Error())
+		s.RespondError(logger, w, srv.ErrInternalServerError)
```
furkhat: What if err = `sql.ErrNoRows`?

tc/tc_linux.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+func classID(ip net.IP) string {
+	return fmt.Sprintf("1:%x", uint16(crc32.ChecksumIEEE(ip)))
+}
+
+func withMask(ip net.IP) string {
+	return fmt.Sprintf("%s/%d", ip, len(ip))
+}
```
furkhat: WOW, this file needs more comments for sure.

sesssrv/start.go
ababo: It will be an internal database inconsistency error.

tc/tc_linux.go
ababo: I'll put some link to a nice manual for linux traffic control: no need to duplicate the info existing in numerous web documents.

PR #289, opened 2018-09-22
Move method to get sessions.

dzeckelev:
Resolves BV-771.


ui/session.go
```diff
@@ -0,0 +1,43 @@
...
+	var (
+		tail string
+		args []interface{}
+	)
+
+	if channel != "" {
+		tail = fmt.Sprintf(
+			"WHERE channel = %s", h.db.Placeholder(1))
+		args = append(args, channel)
+	}
```
furkhat: I really think this should be extracted to a method. Though it's not that big like in another pr.

dzeckelev: Why do we need a method that will be used only once? For decomposition?

furkhat: Code that incapsulates details and show high level logic is good code. You have similar piece of code in [another pr](https://github.com/Privatix/dappctrl/pull/288).

dzeckelev: @furkhat updated.

ui/db.go
```diff
@@ -1,11 +1,57 @@
...
+			typ.Kind().String()).Error(ErrTypeIsNotPointer.Error())
+		return ErrInternal
+	}
+
+	collect := reflect.ValueOf(result).Elem()
+	collectType := collect.Type()
+
+	if collect.Kind() != reflect.Slice {
+		logger.Add("type", collect.Kind().String()).Error(
+			ErrTypeIsNotSlice.Error())
```
furkhat: just log some message. No need to define an error if we return `ErrInternal`.

ui/errors.go
```diff
@@ -31,6 +31,8 @@ const (
 	ErrPrivateKeyNotFound
 	ErrTokenAmountTooSmall
 	ErrBadDestination
+	ErrTypeIsNotPointer
+	ErrTypeIsNotSlice
```
furkhat: Errors listed here are for end users. `ErrTypeIsNotPointer` and `ErrTypeIsNotSlice` are not.

ui/session_test.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"testing"
+
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/ui"
+	"github.com/privatix/dappctrl/util"
+)
+
+func TestGetSessions(t *testing.T) {
+	fxt, assertMatchErr := newTest(t, "GetSessions")
```
furkhat: `assertErrEquals` - just to follow the style of other tests.

ui/db.go
```diff
@@ -1,11 +1,57 @@
...
+func (h *Handler) tailElements(conditions []string) (elements []string) {
+	for k, v := range conditions {
+		condition := fmt.Sprintf(
+			"%s = %s", v, h.db.Placeholder(k+1))
+		elements = append(elements, condition)
+	}
+	return elements
+}
+
+func (h *Handler) getItems(logger log.Logger, view reform.View,
```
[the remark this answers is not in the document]
ababo: I would still argue we don't need this function. More harm then shortening.

ui/db.go
```diff
@@ -1,11 +1,57 @@
...
+	result interface{}, tail string, args ...interface{}) error {
+	objects, err := h.selectAllFrom(logger, view, tail, args...)
+	if err != nil {
+		return err
+	}
+
+	typ := reflect.TypeOf(result)
+
+	if typ.Kind() != reflect.Ptr {
```
ababo: When will we get this condition to executed? The same question to the second condition.

ui/db.go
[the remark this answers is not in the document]
dzeckelev: @furkhat please suggest your solution. You raised this issue.

PR #291, opened 2018-09-23
extended list of offering statuses

furkhat:


ui/offering_test.go
```diff
@@ -357,88 +358,88 @@ func invalidOfferingsArray(
...
-			data.TestPassword, fxt.Offering.ID, action, 100)
-		assertMatchErr(nil, err)
-
-		if j == nil || j.Type != jobType ||
-			j.RelatedID != fxt.Offering.ID {
-			t.Fatal("expected job not created")
-		}
-	}
-}
+// func TestCreateOffering(t *testing.T) {
```
ababo: Why do we need that commented out? This block doesn't explain a thing, so we don't need it since we have a version control.

furkhat: pushed by mistake.

ui/offering_test.go
```diff
@@ -4,12 +4,11 @@ import (
 	"encoding/json"
 	"testing"
 
-	"gopkg.in/reform.v1"
-
 	"github.com/privatix/dappctrl/data"
 	"github.com/privatix/dappctrl/job"
 	"github.com/privatix/dappctrl/ui"
 	"github.com/privatix/dappctrl/util"
+	reform "gopkg.in/reform.v1"
```
dzeckelev: fix import

PR #292, opened 2018-09-24
commands to downgrade the database schema and display version

ubozov:
This commit includes the following changes:

1.  `db-version` - command to display version of the database schema
2.  executing migration scripts to upgrade/downgrade database schema.

Resolves BV-858.

### DIFF
--- a/data/command.go
+++ b/data/command.go
@@ -13,13 +13,10 @@ import (
 	reform "gopkg.in/reform.v1"
 )
 
-type cmdFlag int
-
-// cmdFlags enum
-const (
-	ConnectionString cmdFlag = iota
-	Version
-)
+type cmdFlag struct {
+	connection string
+	version    int64
+}
 
 // ExecuteCommand executes commands to manage database
 // db-migrate - command to execute migration scripts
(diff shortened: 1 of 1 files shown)


data/command.go
```diff
@@ -13,39 +14,58 @@ import (
...
 
 	if *connStr == "" {
 		panic(errors.New("connection string is not detected"))
 	}
-	return *connStr
+
+	m := make(map[string]string)
+
+	m["conn"] = *connStr
+	m["version"] = strconv.Itoa(*version)
```
drew2a: The same as in previous review. 
From my perspective, if some string has been used more than 1 time, then better to extract it to const.

data/command.go
```diff
@@ -5,47 +5,75 @@ import (
...
 		return nil
 	}
 
 	switch args[0] {
 	case "db-migrate":
-		connStr := getConnectionString(args)
-		if err := migration.Update(connStr); err != nil {
+		m := readFlags(args)
+		connStr := m[ConnectionString]
+		version, _ := strconv.ParseInt(m[Version], 10, 64)
```
furkhat: lets not ignore the error.

data/command.go
```diff
@@ -5,47 +5,75 @@ import (
...
 	reform "gopkg.in/reform.v1"
 )
 
+type cmdFlag int
+
+// cmdFlags enum
+const (
+	ConnectionString cmdFlag = iota
+	Version
+)
```
furkhat: what if we define a struct and `readFlags` will return the instance of this struct? It gives better abstraction and encapsulation. 
This way we don't need those constants and type above.

data/command.go

data/command.go
ubozov: good idea, updated

PR #293, opened 2018-09-24
Fix tests with timeout.

dzeckelev:
Resolves BV-829.

### DIFF
--- a/client/bill/monitor_test.go
+++ b/client/bill/monitor_test.go
@@ -5,6 +5,7 @@ package bill
 import (
 	"fmt"
 	"os"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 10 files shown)


client/bill/monitor_test.go
```diff
@@ -80,6 +66,47 @@ func newFixture(t *testing.T, db *reform.DB) *data.TestFixture {
...
+}
+
+func awaitingGoodPosting(wg *sync.WaitGroup, postErrors chan error) {
+	defer wg.Done()
+	for {
+		select {
+		case e := <-postErrors:
+			if e == nil {
+				return
+			}
```
furkhat: is this check redundant?

dzeckelev: no. If the error has become is nil, then you need to complete the function.

furkhat: ok, but it will exit anyway.

dzeckelev: Here an infinite loop. The loop will end either by timeout, or if a error is nul

PR #295, opened 2018-09-26
Checking offering country by IP address from endpoint message.

dzeckelev:
Changes:
* Move country detection functionality to a new package.
* Add the functional definition of a country for any ip address.
* Check offering country by IP address from endpoint message during ClientPreEndpointMsgSOMCGet job  execution.
* Fill `DB.Offering.Country` from `DB.Product.Country`.
* Fill `DB.Product.Country` during pushing product configuration.

Resolves BV-728.

### DIFF
--- a/country/country.go
+++ b/country/country.go
@@ -1,0 +1,77 @@
+package country
+
+import (
+	"context"
+	"encoding/json"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// UndefinedCountry is default country code.
+// This means that a country is not defined.
+const UndefinedCountry = "ZZ"
+
+// Config is the configuration for obtaining a country code.
+type Config struct {
+	Field   string
+	Timeout uint64 // in seconds.
+	// In a url template there should be an pattern {{ip}}.
+	// Pattern {{ip}} will be replaced by a ip address of the agent.
+	URLTemplate string
+}
+
+// NewConfig creates new configuration for obtaining a country code.
+func NewConfig() *Config {
+	return &Config{
+		Timeout: 30,
+	}
+}
+
+func do(ctx context.Context, req *http.Request) (*http.Response, error) {
+	res, err := http.DefaultClient.Do(req.WithContext(ctx))
+	if err != nil {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		default:
+			return nil, err
+		}
+	}
+	return res, nil
+}
+
+// GetCountry returns country code by ip.
+// Parses response in JSON format and returns a value of the field.
+func GetCountry(timeout uint64, url, field string) (string, error) {
+	ctx, cancel := context.WithTimeout(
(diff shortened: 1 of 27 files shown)


proc/worker/country/country.go
```diff
@@ -0,0 +1,96 @@
...
+
+// NewConfig creates new configuration for obtaining a country code.
+func NewConfig() *Config {
+	return &Config{
+		Timeout: 30,
+	}
+}
+
+func do(ctx context.Context, req *http.Request) (*http.Response, error) {
+	res, err := http.DefaultClient.Do(req.WithContext(ctx))
```
ababo: I guess using `http.Client. Timeout` would be easier.

data/schema.sql
```diff
@@ -281,7 +284,8 @@ CREATE TABLE endpoints (
     username varchar(100),
     password varchar(48),
     additional_params json, -- all additional parameters stored as JSON
-    raw_msg text NOT NULL -- raw message in base64 (RFC-4648)
+    raw_msg text NOT NULL, -- raw message in base64 (RFC-4648)
+    country_status country_status_type -- result of checking a country by agent`s ip address.
```
ababo: Why do we need status? I think including detected country would be easier and more informal.

data/schema.sql
```diff
@@ -18,6 +18,9 @@ CREATE TYPE contract_type AS ENUM ('ptc','psc');
 -- Client identification types.
 CREATE TYPE client_ident_type AS ENUM ('by_channel_id');
 
+-- Country status types.
```
ubozov: I think we should reflect these changes additionally in the dbmigration scripts (`data/migration/scripts/00001_schema_down.sql` and `data/migration/scripts/00001_schema_up.sql`).

data/schema.sql
dzeckelev: The offer is not bad, but we agreed on such an implementation.

country/test.go
```diff
@@ -0,0 +1,27 @@
+package country
```
furkhat: I don't like the `<package>/test.go` files. One day we will get rid of them. For now, please add a build tag.
```
// +build !notest
package country
```

country/country.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+// DefaultCountry it is default country code.
+// This means that a country is not defined.
+const DefaultCountry = "ZZ"
+
+// Config is the configuration for obtaining a country code.
+type Config struct {
+	Field   string
+	Timeout time.Duration // in seconds.
```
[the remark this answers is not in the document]
dzeckelev: It is necessary to not make a conversion in this place.
https://github.com/Privatix/dappctrl/blob/945204362c834062475fc3a7fc6da888573e911f/country/country.go#L54

it is convenient to declare a `Timeout` variable in the required format. I wrote a comment specifically

PR #298, opened 2018-10-01
Clearly register error in log about failed authentication.

dzeckelev:
If username is wrong in dappvpn.config.json, then will be errors in a log:
Agent:
* `dappctrl`: WARNING access denied map[sender:ip:port type:sesssrv.Server method:RequireBasicAuth]

Client:
* `dappctrl`: WARNING access denied map[type:sesssrv.Server method:RequireBasicAuth sender:ip:port]
* `dappvpn`: FATAL failed to prepare client configuration: server responed with error: access denied (3) map[channel:channel id]

Resolves BV-880.

### DIFF
--- a/svc/dappvpn/prepare/errors.go
+++ b/svc/dappvpn/prepare/errors.go
@@ -7,13 +7,11 @@ import (
 // Errors.
 const (
 	// CRC16("github.com/privatix/dappctrl/svc/dappvpn/prepare") = 0x23BD
-	ErrGetEndpoint errors.Error = 0x23BD<<8 + iota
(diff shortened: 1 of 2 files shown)


dzeckelev:
@sofabeat If you meant a user name when connecting the client to the agent, then this information is not logged correctly, because a new instance of a `dappvpn` is started. `dappvpn` is used as an authorization script. The log is written in stderr. We need to implement the function of writing a log to a file.

If a username in `dappvpn.config.json` configuration file is not correct, then we will see the errors that I described above.

PR #299, opened 2018-10-01
Validate offering using template.

dzeckelev:
Checks:
1. Check hash match to that in registered in blockchain.
2. Validate offering JSON compliant with offering template JSON.

Resolves BV-888.

### DIFF
--- a/proc/worker/client.go
+++ b/proc/worker/client.go
@@ -1038,7 +1038,7 @@ func (w *Worker) fillOfferingFromSOMCReply(logger log.Logger,
 
 	hashBytes := common.BytesToHash(crypto.Keccak256(offeringData.Offering))
 
-	// check hash match to that in registered in blockchain.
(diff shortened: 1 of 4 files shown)


proc/worker/client.go
```diff
@@ -1036,6 +1036,20 @@ func (w *Worker) fillOfferingFromSOMCReply(logger log.Logger,
 		return nil, ErrOfferingExists
 	}
 
+	hashBytes := common.BytesToHash(crypto.Keccak256(offeringData.Offering))
+
+	// check hash match to that in registered in blockchain.
```
furkhat: start comments with capital letter.

dzeckelev: This is not necessarily inside the function, but I did as you suggest.

dzeckelev: @furkhat updated.

PR #303, opened 2018-10-07
Add auto pop up offerings.

dzeckelev:
Changes:
1) Add `DB.offering.auto_pop_up`.
2) Add `offering.autopopup` setting.
3) Move `EthBackend` from `proc/worker` to `proc/adapter` pkg.
4) Add `proc/looper` pkg for to generate and add jobs on schedule.
5) Add `AutoOfferingPopUp` function to generate `agentPreOfferingPopUp` jobs.

Resolves BV-897.


proc/looper/offering.go
```diff
@@ -0,0 +1,197 @@
...
+
+	"github.com/privatix/dappctrl/data"
+	"github.com/privatix/dappctrl/proc/adapter"
+	"github.com/privatix/dappctrl/util/log"
+)
+
+// AutoOfferingPopUp creates AgentPreOfferingPopUp jobs for active offerings.
+// Taken into account a current balance of agent ETH. If ETH is not enough
+// to pop upped all offerings, then not one of its is pop upped. Function
+// calculates the pop up time.
```
furkhat: `..is not enough to pop up all offerings, then no one of them is popped up..`

proc/looper/offering.go
```diff
@@ -0,0 +1,197 @@
...
+	logger.Debug("started AutoOfferingPopUp")
+	do, err := data.ReadBoolSetting(
+		db.Querier, data.SettingOfferingAutoPopUp)
+	if err != nil {
+		logger.Warn(err.Error())
+	}
+
+	var jobs []*data.Job
+
+	if do {
```
furkhat: To eliminate redundant work and make the use of setting more transparent I suggest to move checking `offering.autopopup` upper.

main.go
```diff
@@ -165,6 +173,39 @@ func createUIServer(conf *ui.Config, logger log.Logger, db *reform.DB,
...
+	if err != nil {
+		return err
+	}
+
+	autoPopUpOfferingFunc := func() []*data.Job {
+		return looper.AutoOfferingPopUp(
+			logger, serviceContractABI, db, ethBack, time.Now)
+	}
+
+	looper.Loop(ctx, logger, db, queue, timeout, autoPopUpOfferingFunc)
```
furkhat: no need to loop if `offering.autopopup` is false.

dzeckelev: What if I want to turn on auto popup later?
There is no need to check the `offering.autopopup` setting.

proc/looper/offering.go
furkhat: Yes, you are right. Didn't think about it.

proc/looper/offering.go
dzeckelev: @furkhat updated.

PR #312, opened 2018-10-12
Update account state.

dzeckelev:
Resolves BV-948.

### DIFF
--- a/doc/ui/rpc.md
+++ b/doc/ui/rpc.md
@@ -184,6 +184,36 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_transfer
 ```
 </details>
 
+#### Update account
+
+*Method*:	`updateAccount`
+
+*Description*: Updates an account.
+
+*Parameters*:
+1. Password (string)
+2. Account id (string)
+3. Name (string)
+4. IsDefault (bool)
+5. inUse (bool)
+
+*Result*: None.
+
+<details><summary>Example</summary>
+    
+```js
+// Request
+curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_updateAccount", "params": ["qwert", "e66d8abd-c5e4-4ced-b9c3-fc3d61a911d0", "user", true, true], "id": 67}' http://localhost:8888/http
+
+// Result
+{
+    "id": 67,
+    "jsonrpc": "2.0",
+    "result": null
+}
+```
+</details>
+
 #### Update balance
 
 *Method*:	`updateBalance`
--- a/ui/account.go
+++ b/ui/account.go
@@ -309,3 +309,30 @@ func (h *Handler) UpdateBalance(password, account string) error {
 
 	return nil
 }
+
+// UpdateAccount updates an account.
+func (h *Handler) UpdateAccount(password, account, name string,
+	isDefault, inUse bool) error {
+	logger := h.logger.Add("method", "UpdateAccount",
+		"account", account)
+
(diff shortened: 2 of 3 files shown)


doc/ui/rpc.md
```diff
@@ -184,6 +184,36 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_transfer
 ```
 </details>
 
+#### Update account status
+
+*Method*:	`updateAccountStatus`
+
+*Description*: Update account state.
```
furkhat: Suggest to leave the word "state". Just `*Description*: Updates an account.`

ui/account.go
```diff
@@ -309,3 +309,30 @@ func (h *Handler) UpdateBalance(password, account string) error {
 
 	return nil
 }
+
+// UpdateAccountStatus updates account state.
+func (h *Handler) UpdateAccountStatus(password, account, name string,
```
furkhat: Lets name it simply `UpdateAccount`, because it seems like it. Right?

ui/account_test.go
```diff
@@ -353,3 +353,43 @@ func TestUpdateBalance(t *testing.T) {
...
+	}
+
+	for name, exp := range testNames {
+		res = handler.UpdateAccountStatus(
+			data.TestPassword, fxt.Account.ID, name, false, false)
+		assertMatchErr(nil, res)
+
+		db.Reload(fxt.Account)
+
+		if fxt.Account.Name != exp {
```
furkhat: You could test all fields update within this loop

ui/account.go
furkhat: Accounts name cannot be described as status. There isn't "state" in the name. I still suggest to rename it as above.

doc/ui/rpc.md
dzeckelev: I think that the word "state" is needed here. I cannot update any account field using this method.

ui/account_test.go
```diff
@@ -353,3 +353,43 @@ func TestUpdateBalance(t *testing.T) {
...
+
+	res := handler.UpdateAccountStatus(
+		"wrong-password", fxt.Account.ID, "", false, false)
+	assertMatchErr(ui.ErrAccessDenied, res)
+
+	oldName := fxt.Account.Name
+	newName := util.NewUUID()[:30]
+
+	testNames := map[string]string{
+		"":      oldName,
```
ababo: This looks suspicious.

doc/ui/rpc.md
furkhat: It updates `name`, `in_use` and `is_default` - all of which are account fields. I don't understand what you mean.

ui/account_test.go
```diff
@@ -353,3 +353,43 @@ func TestUpdateBalance(t *testing.T) {
...
+		}
+	}
+
+	res = handler.UpdateAccountStatus(
+		data.TestPassword, fxt.Account.ID, "", true, true)
+	assertMatchErr(nil, res)
+
+	db.Reload(fxt.Account)
+
+	if fxt.Account.InUse != true || fxt.Account.IsDefault != true {
```
ababo: Why not `!fxt.Account.InUse || !fxt.Account.IsDefault`?

ui/account_test.go
dzeckelev: What confuses you? There is a check, if a name is empty, then it should not be changed.

dzeckelev:
@furkhat @ababo updated.

PR #316, opened 2018-10-15
Make the client/svcrun package to support Windows.

ababo:
This PR includes the following changes:
1. The package has been refactored to be platform-agnostic.
2. Added a new client/svcrun/exec package with platform-specific
   implementation for Unix and Windows.

Partially resolves BV-921.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 .DS_Store
 
 *_reform.go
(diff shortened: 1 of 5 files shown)


ababo:
@sofabeat That's because of Go, in C using native WinAPI it's much simpler.

PR #319, opened 2018-10-16
Fix dappctrl windows build

sofabeat:
Fix statik go generate issues


statik/misc_windows.go
```diff
@@ -0,0 +1,38 @@
...
+
+	"github.com/rakyll/statik/fs"
+)
+
+//go:generate .\pkgList\gen_win.bat
+//go:generate go run ../tool/copy_dbscripts/copy.go
+//go:generate cmd /C del /F /Q statik.go
+//go:generate statik -f -src=. -dest=..
+
+// File paths.
```
ababo: I think we should put the following content into a dedicated file (shared between different architectures).

statik/pkgList/gen_win.bat
```diff
@@ -0,0 +1 @@
+go list ../... > .\pkgList\packages.txt
```
drew2a: If it not a common `statik` file, may be better to move it to the `scripts` folder?

ababo: What is the purpose of this file?

sofabeat: @dzeckelev @ababo - what do you think?

statik/misc_windows.go
sofabeat: ok

sofabeat:
New pull request #320 is opened instead of this one. Closing it.

PR #337, opened 2018-10-25
Catch panic to file logger.

dzeckelev:
Resolves BV-960.

### DIFF
--- a/main.go
+++ b/main.go
@@ -142,7 +142,7 @@ func createLogger(conf *config, db *reform.DB) (log.Logger, io.Closer, error) {
 		return nil, nil, err
 	}
 
-	logger := log.NewMultiLogger(elog, flog, dlog, blog)
+	logger := log.NewMultiLogger(flog, elog, dlog, blog)
 
 	blog2 := blog.(bugsnag.Log)
 	reporter, err := bugsnag.NewClient(conf.Report, db, blog2,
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -201,6 +201,25 @@ func startAutoPopUpLoop(ctx context.Context, cfg *looper.Config, period uint32,
...
+		case string:
+			if err == "fatal log event" {
+				panic(err)
+			}
+			v = err.(string)
+		default:
+			v = fmt.Sprintf("error %+v", v)
+		}
+		logger.Fatal(v)
+		panic(err)
```
furkhat: this panic is redundant. `logger.Fatal` calls panic.

furkhat: I may be wrong, but this seems like an infinite panic-recover-panic cycle. Did you test it?

dzeckelev: You're wrong. If a panic has already been processed by a logger, then the function ends on this line.
https://github.com/Privatix/dappctrl/blob/4a1c7459ceab7a9eb7c28cc23c5bbe8bbb34613c/main.go#L212
After that, the `defer bugsnag.PanicHunter()` will catch and process the panic.
Yes, I tested it.

dzeckelev: `logger.Fatal(v)`  is necessary to catch an unprocessed panic.

furkhat: why panic on the next line?

dzeckelev: @furkhat this is an extra string. Thank you. Updated.

PR #338, opened 2018-10-26
Fix GetClientChannels.

dzeckelev:
Resolves BV-1012.

### DIFF
--- a/ui/channel.go
+++ b/ui/channel.go
@@ -187,8 +187,19 @@ func (h *Handler) GetClientChannels(password, channelStatus,
 		return nil, err
 	}
 
+	jobFilter := `channels.id IN
+			(SELECT related_id 
+			   FROM jobs
+			  WHERE jobs.type = '%s' AND jobs.status = '%s')`
+
+	jobDoneCondition := fmt.Sprintf(jobFilter,
+		data.JobClientAfterChannelCreate, data.JobDone)
+
+	condition := fmt.Sprintf("%s AND %s", clientChannelsCondition,
+		jobDoneCondition)
+
 	chs, total, err := h.getChannels(logger, channelStatus, serviceStatus,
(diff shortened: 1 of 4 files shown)


ui/channel.go
```diff
@@ -282,26 +291,29 @@ func (h *Handler) createClientChannelResult(logger log.Logger,
...
 
+	job2 := item.(*data.Job)
+
 	logger = logger.Add("job", job2, "offering",
 		offering, "channel", channel)
 
 	if job2.Status != data.JobDone {
-		logger.Warn("job status is not done")
-		return nil, nil
+		logger.Error(ErrJobNotDone.Error())
```
ababo: Why do we need to log the same error as we return? IMO it doesn't make too much sense.

ui/channel.go
```diff
@@ -187,8 +187,19 @@ func (h *Handler) GetClientChannels(password, channelStatus,
 		return nil, err
 	}
 
+	jobFilter := `channels.id IN (SELECT related_id
```
ababo: Better would be
```go
channels.id IN
  (SELECT related_id 
     FROM jobs
    WHERE jobs.type = '%s' AND jobs.status = '%s')
```

ui/db.go
```diff
@@ -61,6 +61,20 @@ func (h *Handler) selectAllFrom(logger log.Logger, view reform.View,
...
+	args ...interface{}) (reform.Struct, error) {
+	items, err := h.db.SelectOneFrom(view, tail, args...)
+	if err != nil {
+		logger.Error(err.Error())
+		if err == reform.ErrNoRows {
+			return nil, notFoundError
+		}
+		return nil, ErrInternal
+	}
+	return items, err
```
ababo: better `return items, nil`.

PR #345, opened 2018-11-03
Get offerings filter parameters for client.

dzeckelev:
Resolves BV-1022.

### DIFF
--- a/ui/offering.go
+++ b/ui/offering.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
+
 	"github.com/privatix/dappctrl/data"
 	"github.com/privatix/dappctrl/job"
 	"github.com/privatix/dappctrl/messages"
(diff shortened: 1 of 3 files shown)


ui/offering.go
```diff
@@ -428,3 +440,70 @@ func (h *Handler) CreateOffering(password string,
...
+		var country string
+		var count int64
+
+		if err := rows.Scan(&country, &count); err != nil {
+			logger.Error(err.Error())
+			return nil, ErrInternal
+		}
+		countries = append(countries, country)
+	}
+	return countries, nil
```
ababo:
```
if err := rows.Err(); err != nil {
	log.Fatal(err)
}
```

ui/offering.go
```diff
@@ -428,3 +440,75 @@ func (h *Handler) CreateOffering(password string,
...
+
+		if err := rows.Scan(&country, &count); err != nil {
+			logger.Error(err.Error())
+			return nil, ErrInternal
+		}
+		countries = append(countries, country)
+	}
+
+	if err := rows.Err(); err != nil {
+		logger.Fatal(err.Error())
```
ababo: Sorry, I didn't mean to literally put `Fatal`. Should be `Error()` and to be returned.

PR #353, opened 2018-11-13
create db via dappctrl db-create command

ubozov:
Resolve BV-1027.

### DIFF
--- a/data/command.go
+++ b/data/command.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"regexp"
 	"strings"
 
 	"github.com/privatix/dappctrl/data/migration"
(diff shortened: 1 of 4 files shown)


data/command.go
```diff
@@ -27,6 +30,12 @@ func ExecuteCommand(args []string) error {
 	}
 
 	switch args[0] {
+	case "db-create":
+		f := readFlags(args)
+		if err := createDatabase(f.connection); err != nil {
+			panic(fmt.Sprintf("failed to create database %s", err))
```
ababo: I would write `panic("failed to create database: " + err.Error())`.

data/command.go
```diff
@@ -94,3 +103,50 @@ func initData(connStr string) error {
 		return nil
 	})
 }
+
+func createDatabase(connStr string) error {
+	db, err := sql.Open("postgres", connStr)
```
ababo: Please reuse routines in the `data`-package.

data/command.go

data/command.go

PR #363, opened 2018-11-21
Replace status strings to status arrays in a getAgentChannels method.

dzeckelev:
Resolves BV-1133.

### DIFF
--- a/doc/ui/rpc.md
+++ b/doc/ui/rpc.md
@@ -354,8 +354,8 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_changeCh
 
 *Parameters*: 
 1. Password (string)
-2. Channel status (string)
-3. Service status (string)
+2. Channel statuses (array of `string`s, can be empty)
+3. Service statuses (array of `string`s, can be empty)
 4. Offset (number)
 5. Limit (number)
 
@@ -367,7 +367,7 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_changeCh
 
 ```js
 // Request
-curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_getAgentChannels", "params": ["qwerty","active","pending",0,1], "id": 67}' http://localhost:8888/http
+curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_getAgentChannels", "params": ["qwerty",["active"],["pending", "terminated"],0,1], "id": 67}' http://localhost:8888/http
 
 // Result
 {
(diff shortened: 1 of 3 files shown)


ui/channel.go
```diff
@@ -187,8 +187,17 @@ func (h *Handler) GetClientChannels(password, channelStatus,
 	if err := h.checkPassword(logger, password); err != nil {
 		return nil, err
 	}
+	// TODO(maxim): Replace channelStatus, serviceStatus arguments types from string to []string.
```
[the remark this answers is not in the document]
furkhat: resolved?

ui/channel.go
```diff
@@ -342,30 +351,44 @@ func (h *Handler) createClientChannelResult(logger log.Logger,
...
+		condition := statusCondition(channelStatuses, "channel_status")
+		conditions = append(conditions, condition)
+		for _, channelStatus := range channelStatuses {
+			args = append(args, channelStatus)
+		}
 	}
 
-	items := h.tailElements(conditions)
+	if len(serviceStatuses) != 0 {
+		condition := statusCondition(serviceStatuses, "service_status")
```
furkhat: a little duplication. Can we extract whole `if` with its body to a method, and just call the method twice? Smth like this:
```
args = append(method(channelStatuses, "channel_status"), args...)
args = append(method(serviceStatuses, "service_status"), args...)
```

ui/channel.go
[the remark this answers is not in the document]
dzeckelev: Now it is not necessary.

dzeckelev:
@furkhat updated.

doc/ui/rpc.md
```diff
@@ -354,8 +354,8 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_changeCh
 
 *Parameters*: 
 1. Password (string)
-2. Channel status (string)
-3. Service status (string)
+2. Channel statuses (array of `string`s or empty array)
```
ababo: I would write "array of `string`s, can be empty".

ui/channel.go
```diff
@@ -187,8 +187,17 @@ func (h *Handler) GetClientChannels(password, channelStatus,
 	if err := h.checkPassword(logger, password); err != nil {
 		return nil, err
 	}
+	// TODO(maxim): Replace channelStatus, serviceStatus arguments types from string to []string.
+	chStatuses := make([]string, 0)
```
ababo: `var chStatuses []string` of `chStatuses:=[]string{}` is shorter, up to you.

PR #367, opened 2018-11-22
add tor setting

furkhat:
Add tor setting to dynamically control what somc type to use.
`AgentPreOfferingMsgBCPublish` and `AgentPreOfferingPopUp` - based on setting decides what somc type to use on publish.
`AgentPreEndpointMsgCreate` and `AgentAfterOfferingMsgBCPublish` - based on related offering's somc type decides whether to publish endpoint or not.

### DIFF
--- a/data/settings.go
+++ b/data/settings.go
@@ -18,6 +18,7 @@ const (
 	SettingMinConfirmations   = "eth.min.confirmations"
 	SettingPasswordHash       = "system.password"
 	SettingPasswordSalt       = "system.salt"
+	SettingSOMCUseTor         = "somc.transport.tor"
 )
 
 // ReadSetting reads value of a given setting.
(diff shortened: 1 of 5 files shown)


data/settings.go
```diff
@@ -18,6 +18,7 @@ const (
 	SettingMinConfirmations   = "eth.min.confirmations"
 	SettingPasswordHash       = "system.password"
 	SettingPasswordSalt       = "system.salt"
+	SettingSOMCUseTor         = "somc.transport.tor"
```
ababo: What about having `SettingSOMCTransport` with predefined set of values ("server", "tor", etc.)?

furkhat: I think it depends on UX. We don't have validation of settings changes, and if UI displays this setting as editable text field user will be able to type "Tor" instead of "tor", which might lead to some bizarre behaviour.

I will check UI, if it can do dropdowns (or similar), I will change the code to work with predefined set of values.

furkhat: No validation. UI presents just text field. For now leaving as it is.

PR #397, opened 2018-12-10
tor is the only somc transport

furkhat:
Removed `status` field from enpoints table.
Removed `msg_channel_published` status and made related changes to dappctrl, ui and initializer.py needs to be updated.
Removed centrelised somc and related jobs.
Added ping to somc before accepting an offering.


proc/worker/agent_test.go
```diff
@@ -339,18 +336,18 @@ func TestAgentPreEndpointMsgCreate(t *testing.T) {
...
 
-	runJob(t, env.worker.AgentPreEndpointMsgCreate, fixture.job)
+	runJob(t, env.worker.AgentPreEndpointMsgCreate, fxt.job)
 
 	endpoint := &data.Endpoint{}
 	if err := env.db.SelectOneTo(endpoint,
-		"where template=$1 and channel=$2 and status=$3",
-		fixture.TemplateAccess.ID,
-		fixture.Channel.ID, data.MsgUnpublished); err != nil {
+		"where template=$1 and channel=$2 and id<>$3",
```
ababo: WHERE, AND

proc/worker/client.go
```diff
@@ -991,47 +971,29 @@ func (w *Worker) clientRetrieveAndSaveOffering(logger log.Logger,
 	job *data.Job, block uint64, somcType uint8, somcData data.Base64String,
 	agentAddr common.Address, hash common.Hash, currentSupply uint16) error {
```
ababo: Why don't we add args to logging context?

PR #398, opened 2018-12-11
Rewrite the session server to support RPC

ababo:
Resolves BV-972


sess/handler_test.go
```diff
@@ -0,0 +1,61 @@
+// +build !nosesstest
```
furkhat: no need in build tag for *_test.go files.

sess/product_test.go
```diff
@@ -0,0 +1,116 @@
+// +build !nosesstest
```
furkhat: same here

sess/handler_test.go
ababo: The purpose of this tag is to exclude specific tests from test launches. Not sure how this is still relevant (long time ago we agreed on that to be able to exclude `eth` tests).

sofabeat:
Con you, please remove SOMC in "resolve conflicts". It is gone. No more external SOMC, only internal SOMC server.

ababo:
@sofabeat conflicts resolved

PR #405, opened 2018-12-14
Recover service statuses for client channels

ababo:
Partially resolves BV-1225

### DIFF
--- a/data/account_test.go
+++ b/data/account_test.go
@@ -4,21 +4,12 @@ import (
 	"bytes"
 	"crypto/ecdsa"
 	"crypto/rand"
-	"os"
 	"testing"
 
 	"github.com/ethereum/go-ethereum/crypto"
 
 	"github.com/privatix/dappctrl/data"
-	"github.com/privatix/dappctrl/util"
 )
-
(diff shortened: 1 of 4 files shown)


data/account_test.go
```diff
@@ -1,25 +1,18 @@
+// +build !nodatatest
```
furkhat: redundant build tag, *_test files ignored by default.

data/recover.go
```diff
@@ -0,0 +1,40 @@
+package data
+
+import (
+	"fmt"
+
+	"gopkg.in/reform.v1"
+)
+
+// Recover ensures data consistency after unexpected controller's exit.
```
furkhat: Please extend this docstring. Write how you do it
```
Recover ensures data consistency after unexpected controller's exit by moving all semi active services to suspended state```
```

ababo: @furkhat updated

data/account_test.go
ababo: @furkhat Will discuss that with the team.

ababo: @furkhat removed them

PR #420, opened 2018-12-27
ui uses token

furkhat:


doc/ui/rpc.md
```diff
@@ -260,6 +260,32 @@ curl -X POST -H "Content-Type: application/json" --data '{"method": "ui_generate
 
 ### Authentication
 
+#### Get Token
+
+*Method*: `getToken`
+
+*Description*: Given correct password, generates and returns new access token.
```
ababo: IMO we need to clarify if it's per session, or is it going to expire some time.

ui/handler.go
```diff
@@ -21,6 +21,7 @@ type Handler struct {
 	userRole          string
 	processor         *proc.Processor
 	somcClientBuilder somc.ClientBuilderInterface
+	token             TokenMakeChecker
```
ababo: So we have a single token per all sessions? What if we'll have multiple sessions (it's totally impossible)?

furkhat: Yes, for now we are going with simples possible solution.

PR #431, opened 2019-01-28
initializer BV-224

velopokatun:


drew2a:
Resolves privatix/privatix#74

PR #436, opened 2019-03-07
validate tpl hash

furkhat:


proc/worker/client.go
```diff
@@ -341,21 +340,20 @@ func (w *Worker) extractEndpointMessage(logger log.Logger,
...
+	if err != nil {
+		return nil, err
 	}
 
-	var msg ept.Message
-	err = json.Unmarshal(mdata, &msg)
-	if err != nil {
-		logger.Error(err.Error())
-		return nil, ErrInternal
+	if !util.ValidateJSON(tpl.Raw, mdata) {
```
sofabeat: Error handling here is unnecessary?

furkhat: No. Unmarshal does not validate skipping of fields required in schema.

PR #437, opened 2019-03-14
create sess.Client

furkhat:


sofabeat:
Do we leave possibility to write any 3rd party adapter on any language? Do introduced changes simplifies it?

PR #447, opened 2019-05-06
optional trace,cpu,mem profiling

furkhat:


furkhat:
@sofabeat added cpu,mem,trace profiles all together.

PR #461, opened 2019-07-05
changed initial rating values expression & return max rating to ui

furkhat:


proc/worker/client.go
```diff
@@ -1303,18 +1303,20 @@ func (w *Worker) closingEvents(logger log.Logger) ([]closingEvent, error) {
...
 		}
 	}
-	// Avoid devision by zerro.
-	if total == 0 {
-		return 0
-	}
 
-	return uint64(success / total * 10e9)
+	// geometric_progression_sum * ratio_of_coops_over_uncoops
+	return uint64((2*amount - 2*amount/float64(uint(2)<<(uint(len(events))-1))) * nCoop / float64(len(events)))
```
drew2a: It looks over complex. My suggestion: to split.

PR #462, opened 2019-07-08
update gas limit, autopopup, error reporting

sofabeat:
- some gas limits were completely removed, as they apply to public smart-contract method, which doesn't require gas at all
- error reporting was disabled by default
- gas limit updated to be closer to real usage, while still having 10% to 30% margin

### DIFF
--- a/dappctrl-dev.config.json
+++ b/dappctrl-dev.config.json
@@ -63,24 +63,19 @@
     },
     "Gas": {
         "PSC": {
-            "AddBalanceERC20": 100000,
-            "BalanceOf": 100000,
-            "CooperativeClose": 200000,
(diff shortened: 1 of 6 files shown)


dappctrl-dev.config.json
```diff
@@ -186,7 +181,7 @@
         "Workers": 0
     },
     "Looper": {
-        "AutoOfferingPopUpTimeout": 172800000
+        "AutoOfferingPopUpTimeout": 86400000
```
sofabeat: @drew2a - here is popup frequency ~ 24 hours

drew2a: ok, thanks

PR #474, opened 2019-08-14
query new events from the latest block, search for offerings in old blocks

furkhat:


furkhat:
@sofabeat No need to create migration for prod data.

PR #477, opened 2019-08-16
long setting name, popups processed not in order

furkhat:


proc/worker/client.go
```diff
@@ -1051,7 +1056,8 @@ func (w *Worker) fillOfferingFromMsg(logger log.Logger, offering []byte,
 	logger = logger.Add("offering", offering)
 	_, err := w.offeringByHashString(logger, hash)
 	if err == nil {
-		return nil, ErrOfferingExists
+		logger.Warn("offerings already exists")
```
drew2a: Maybe it's better to add what the particular offering it is

PR #479, opened 2019-08-19
do not run back monitoring after its done. prioritise jobs

furkhat:


job/queue.go
```diff
@@ -260,7 +260,7 @@ func (q *queue) processMain() error {
 			  SELECT DISTINCT ON (related_id) *
 			    FROM jobs
 			   WHERE status = $1
-			   ORDER BY related_id, created_at) AS ordered
+			   ORDER BY related_id, related_type, created_at) AS ordered
```
sofabeat: I don't see the case you need it. `related_id` is `uuid` and should be unique.

furkhat: You're right. I had smth different in mind.

job/queue.go
```diff
@@ -262,6 +262,7 @@ func (q *queue) processMain() error {
 			   WHERE status = $1
 			   ORDER BY related_id, created_at) AS ordered
 			 WHERE not_before <= $2
+			 ORDER BY related_type
```
furkhat: The idea is to prioritize channel and endpoint related jobs over offering related jobs. With this order should be like `account, channel, endpoint, offering`.

